PR #15, opened 2021-03-04
feat: add default schema name

IlyaFaer:


IlyaFaer:
The checks here are failed because of those INSERT error Hemang is fixing now. As the checks are using `python-spanner` from `master` branch, they don't have a fix/revert yet. The screenshot I've attached is from my local env, where the erroring PR is reverted.

larkee:
@IlyaFaer Looking at the failing tests in the check, these two seem to be consistently failing:
```
FAILED test/test_suite.py::EscapingTest::test_percent_sign_round_trip - Asser...
FAILED test/test_suite.py::ExceptionTest_spanner+spanner::test_integrity_error
```
I assume these are the tests that are affected by the INSERT bug and will disregard them in my reviews :+1:

IlyaFaer:
@larkee,

> I assume these are the tests that are affected by the INSERT bug

Yes it is

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -171,6 +171,14 @@ def _check_unicode_returns(self, connection, additional_tests=None):
         """Ensure requests are returning Unicode responses."""
         return True
 
+    def _get_default_schema_name(self, _):
+        """Get default Cloud Spanner schema name.
+
+        Returns:
+            str: Schema name.
+        """
+        return "unnamed"
```
larkee, line 180: From [the docs](https://cloud.google.com/spanner/docs/information-schema#schemata):

> The name of the schema. An empty string if unnamed.

Does SQLAlchemy support an empty string schema name?

IlyaFaer: I never saw any restrictions like that. I don't actually think SQLAlchemy API by itself uses schema names. Pushing an empty string.

PR #16, opened 2021-03-04
test: fix the boolean literal test

IlyaFaer:
The test is trying to insert two rows into a table with empty primary key and fails with syntax error. Overriding the test to split insert into two separated operations.

### DIFF
--- a/test/test_suite.py
+++ b/test/test_suite.py
@@ -136,8 +136,8 @@ class BooleanTest(_BooleanTest):
         """
         SPANNER OVERRIDE:
 
-        Cloud Spanner supports tables with empty primary key, but
-        only single one row can be inserted into such a table -
+        Cloud Spanner supports tables with an empty primary key, but
+        only a single row can be inserted into such a table -
         following insertions will fail with `Row [] already exists".
         Overriding the test to avoid the same failure.
         """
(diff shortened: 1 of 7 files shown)


larkee:
`BooleanTest_spanner+spanner.test_render_literal_bool` is failing due to `google.api_core.exceptions.FailedPrecondition: 400 Duplicate name in schema: t.`

IlyaFaer:
> BooleanTest_spanner+spanner.test_render_literal_bool is failing due to google.api_core.exceptions.FailedPrecondition: 400 Duplicate name in schema: t.

Yeah, it's one of those errors, when another test added a table with the same name, but wasn't able to clean it up because of an internal error (not assertion).

I can only show you that locally with PR 223 (the one about INSERT error) reverted, everything is passing fine - not only the overriden case, but all of the boolean tests.

Безымянный

larkee (approved):
LGTM, given that the override logic makes sense and the failure looks related to the `db_api` implementation issue.

test/test_suite.py
```diff
@@ -68,3 +70,17 @@ def test_percent_sign_round_trip(self):
...
             )
+
+
+class BooleanTest(_BooleanTest):
+    def test_render_literal_bool(self):
+        """
+        SPANNER OVERRIDE:
+
+        Cloud Spanner supports tables with empty primary key, but
+        only single one row can be inserted into such a table -
```
larkee, line 81: nit:
```suggestion
        only a single row can be inserted into such a table -
```

PR #18, opened 2021-03-08
feat: add empty sets support

IlyaFaer:


IlyaFaer:
@larkee, implementing a small feature. Most of the dialects do the very same things to get an empty set casted to a particular type.

The PR makes a bunch of tests passing:

Безымянный

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -54,6 +68,20 @@ class SpannerSQLCompiler(SQLCompiler):
...
+
+        Args:
+            type_ (sqlalchemy.sql.sqltypes.SchemaType):
+                A SQLAlchemy data type.
+
+        Returns:
+            str: A query to select an empty set of the given type.
+        """
+        return "SELECT CAST(1 AS {}) FROM (SELECT 1) WHERE 1 != 1".format(
+            _type_map_inv[type(type_[0])]
```
larkee, line 82: I'm not familiar with the `SchemaType` but why isn't it (type(type_)) or something similar?

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -39,6 +39,20 @@
...
 
+_type_map_inv = {
+    types.Boolean: "BOOL",
+    types.BINARY: "BYTES(MAX)",
+    types.DATE: "DATE",
+    types.DATETIME: "DATETIME",
+    types.Float: "FLOAT64",
+    types.BIGINT: "INT64",
+    types.DECIMAL: "NUMERIC",
+    types.String: "STRING",
```
larkee, line 50: If `BYTES` requires the length to be specified, wouldn't `STRING` need the same?

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
IlyaFaer: In fact I can't say anything straight on it. The method is not documented (even in [source code](https://github.com/sqlalchemy/sqlalchemy/blob/1edbe841bcb43d761f5084a2e77e8f37ee041c35/lib/sqlalchemy/sql/compiler.py#L1894)). The argument is a list with an empty value of some type in it. Few dialects are implementing this method, and they don't try to look at the given type, they just return an empty set of `int` type. It's not correct - I've tried to run several queries, mixingthe types, to ensure it gives errors of a kind `can't cast type1 to int64`.

Thus, I'm taking the first (and the only) element of the given list, take a type of it and then see for a correct type name.

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
IlyaFaer: You probably right, I just haven't yet taken a look at `STRING` type tests. The `STRING` name is working fine in other tests (which create tables with strings and insert data into it), so I didn't touch it yet.

larkee: Fair enough. It's fine for now :+1:

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
larkee: I see. I don't like this line from a readability perspective but it seems to work and since the `sqlalchemy.sql.sqltypes.SchemaType` seems to be subscriptable I guess there's no problem

larkee (approved):

PR #19, opened 2021-03-08
fix: escape and exception tests

HemangChothani:


test/test_suite.py
```diff
@@ -68,3 +73,21 @@ def test_percent_sign_round_trip(self):
...
+            conn.execute(self.tables.manual_pk.insert(), {"id": 1, "data": "d1"})
+
+            assert_raises(
+                exc.IntegrityError,
+                conn.execute,
+                self.tables.manual_pk.insert(),
+                {"id": 1, "data": "d1"},
+            )
+
+            trans.rollback()
```
IlyaFaer, line 93: This changes the behavior of the test. It was working fine before the PR 223, more than that, there are at least 3 other tests, which are failing because of 223 PR. Thus, I assume it should be fixed in python-spanner and insert helpers/parsers, not by patching every failing test.

larkee:
Closing this PR because the issue is due to the DBAPI connector so the change will need to be made there.

PR #20, opened 2021-03-08
test: fix integer compliance tests

HemangChothani:


test/test_suite.py
```diff
@@ -68,3 +71,68 @@ def test_percent_sign_round_trip(self):
...
+
+        if util.py3k:
+            assert isinstance(row[0], int)
+        else:
+            assert isinstance(row[0], (long, int))  # noqa
+
+        config.db.execute(int_table.delete())
+
+    @provide_metadata
+    def _literal_round_trip(self, type_, input_, output, filter_=None):
```
larkee, line 103: I'm a little confused about what these two methods are for. Are they generating metadata for the integer compliance tests? Could you explain what these methods are doing/how they are used?

test/test_suite.py
```diff
@@ -420,3 +424,87 @@ class TimeTests(_TimeMicrosecondsTest, _TimeTest):
...
+        """
+        SPANNER OVERRIDE:
+
+        Spanner is not able cleanup data and drop the table correctly,
```
larkee, line 471: Do you know why Spanner is not cleaning up the data and dropping the table correctly? Could #28 potentially resolve this?

This override reason doesn't seem tied to a feature that Spanner doesn't support such as temporary tables or multiple rows in empty primary key tables so I think this may be a bug rather than something we should override.

test/test_suite.py
```diff
@@ -420,3 +424,87 @@ class TimeTests(_TimeMicrosecondsTest, _TimeTest):
...
+        SPANNER OVERRIDE:
+
+        This is the helper method for integer class tests which creates table and
+        perform insert operation.
+        Cloud Spanner supports tables with empty primary key, but only single one
+        row can be inserted into such a table - following insertions will fail with
+        `400 id must not be NULL in table date_table`.
+        Overriding the tests and add a manual primary key value to avoid the same
+        failures and delete the from the table at the end.
+        """
```
larkee, line 442: minor nit:
```
        This is the helper method for integer class tests which creates a table and
        performs an insert operation.
        Cloud Spanner supports tables with an empty primary key, but only one
        row can be inserted into such a table - following insertions will fail with
        `400 id must not be NULL in table date_table`.
        Overriding the tests and adding a manual primary key value to avoid the same
        failures and deleting the table at the end.
```

test/test_suite.py
larkee: Locally, I can get this test to pass without the override by changing the `db_api` implementation to execute DDL statements immediately instead of queuing them until a non-DDL statement is executed.

Please add a TODO to remove this override

HemangChothani: Ok, because  locally implemented both the PRs https://github.com/cloudspannerecosystem/python-spanner-sqlalchemy/pull/28, https://github.com/googleapis/python-spanner/pull/277, but still failing for me.

larkee: The required change is shown below:
```
--- a/google/cloud/spanner_dbapi/cursor.py
+++ b/google/cloud/spanner_dbapi/cursor.py
@@ -174,14 +174,9 @@ class Cursor(object):
         try:
             classification = parse_utils.classify_stmt(sql)
             if classification == parse_utils.STMT_DDL:
-                self.connection._ddl_statements.append(sql)
+                self.connection.database.update_ddl([sql]).result()
                 return
 
-            # For every other operation, we've got to ensure that
-            # any prior DDL statements were run.
-            # self._run_prior_DDL_statements()
-            self.connection.run_prior_DDL_statements()
-
             if not self.connection.autocommit:
                 if classification == parse_utils.STMT_UPDATING:
                     sql = parse_utils.ensure_where_clause(sql)
```

AVaksman: The [docs](https://cloud.google.com/spanner/docs/schema-updates?authuser=2#large-updates) here and some other places do seem to lean towards batching the DDL statements since they might be quite time consuming. That was probably the original idea behind this implementation.
Perhaps we can offer a solution where we would execute DDL statement in `autocommit=True` mode and keep the current implementation in `autocommit=False` mode.
I think it will be intuitive from the user's perspective.

```diff
--- a/google/cloud/spanner_dbapi/cursor.py
+++ b/google/cloud/spanner_dbapi/cursor.py
        try:
             classification = parse_utils.classify_stmt(sql)
             if classification == parse_utils.STMT_DDL:
                 self.connection._ddl_statements.append(sql)
+                if self.connection.autocommit:
+                    self.connection.run_prior_DDL_statements()
                 return
 
            # For every other operation, we've got to ensure that
            # any prior DDL statements were run.
            # self._run_prior_DDL_statements()
            self.connection.run_prior_DDL_statements()

            if not self.connection.autocommit:
                if classification == parse_utils.STMT_UPDATING:
                    sql = parse_utils.ensure_where_clause(sql)
```
@larkee WDYT?

HemangChothani: @AVaksman Your given code snippet is failing the tests. Earlier test was passed with @larkee 's code snippet.

larkee: @AVaksman I think this will make sense as long as we clearly document that for `autocommit=False` DDL statements are not executed until a non-DDL statement is executed. We can suggest that for cases where only DDL statements need to be executed that they are followed by "SELECT 1" e.g.
```
conn.execute(create_ddl_stmt)
conn.execute(update_ddl_stmt)
conn.execute("SELECT 1")
```

@HemangChothani This is likely because `autocommit` is `False` by default (I assume?). See the above sample for an idea of how we may need to override the tests.

Another thing to note here is that unlike other databases, Spanner does not support running DDL statements in a transaction which means DDL statements cannot be rolled back. There is no fix for this and so this will have to be clearly documented.

HemangChothani: @larkee Yes, I was talking about `autocommit=Flase`. I think need to remove TODO note then.

larkee: Correct. Please remove it and update the override explanation to mention that DDL statements aren't executed until a non-DDL statement is executed :+1:

HemangChothani: @larkee  Still this doesn't work for me with changes recommended by alex, am i missing something here? 
```python

def _literal_round_trip(self, type_, input_, output, filter_=None):
    t = Table("int_t", self.metadata, Column("x", type_))
        t.create()

        with db.connect() as conn:
            for value in input_:
                ins = (
                    t.insert()
                    .values(x=literal(value))
                    .compile(
                        dialect=db.dialect, compile_kwargs=dict(literal_binds=True),
                    )
                )
                conn.execute(ins) --> thorws an error
                conn.execute("SELECT 1")  # didn't execute
            if self.supports_whereclause:
                stmt = t.select().where(t.c.x == literal(value))
            else:
                stmt = t.select()

            stmt = stmt.compile(
                dialect=db.dialect, compile_kwargs=dict(literal_binds=True),
            )
            for row in conn.execute(stmt):
                value = row[0]
                if filter_ is not None:
                    value = filter_(value)
                assert value in output

Comment (HemangChothani) [reply test/test_suite.py:474]: @larkee Test passes  for the very first time when table doesn't exist and after that it failed every time with an error `400 Duplicate name in schema: int_t`. 
`DROP TABLE ` is still not working properly after merging all the related branches (with current master branch of spanner and spaner-sqlalchemy). I think the culprit is `autocommit=False` because when set `autocommit=True` in spanner_dbapi everything works fine and table deleted successfully after every test.

Solution might be execute non-DDL  statement after the `DROP TABLE` or commit manually.

Comment (larkee) [reply test/test_suite.py:474]: >Solution might be execute non-DDL statement after the `DROP TABLE`

This is what I would suggest but I'm not sure where `DROP TABLE` is called between tests. For now I think the current implementation is fine. We can revisit this another time :+1:

File: test/test_suite.py [hunk 4]
```diff
@@ -422,3 +425,89 @@ class TimeTests(_TimeMicrosecondsTest, _TimeTest):
...
+
+        SPANNER OVERRIDE:
+
+        Spanner is not able cleanup data and drop the table correctly,
+        table was already exists after related tests finished, so it doesn't
+        create a new table and when started tests for other data type  following
+        insertions will fail with `400 Duplicate name in schema: t.
+        Overriding the tests to create a new table for test and drop table manually
+        before it creates a new table to avoid the same failures.
+        """
```
larkee, line 480: Minor nit:
```
        SPANNER OVERRIDE:

        Spanner DBAPI does not execute DDL statements unless followed by a
        non DDL statement, which is preventing correct table clean up.
        The table already exists after related tests finish, so it doesn't
        create a new table and when running tests for other data types
        insertions will fail with `400 Duplicate name in schema: t`.
        Overriding the tests to create and drop a new table to prevent
        database existence errors.
```

larkee (approved):

PR #21, opened 2021-03-10
test: skip underscore-names tests

IlyaFaer:


google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -130,6 +130,7 @@ class SpannerDialect(DefaultDialect):
     positional = False
     paramstyle = "format"
     encoding = "utf-8"
+    max_identifier_length = 127
```
IlyaFaer, line 133: The [Spanner docs](https://cloud.google.com/spanner/docs/data-definition-language#naming_conventions) says the length is *between*, so I assume it means the max length is 127

test/test_suite.py
```diff
@@ -143,3 +142,11 @@ def test_render_literal_bool(self):
...
         self._literal_round_trip(Boolean(), [True], [True])
         self._literal_round_trip(Boolean(), [False], [False])
+
+
+class TableDDLTest(_TableDDLTest):
+    @pytest.mark.skip(
+        "Spanner table name must start with an uppercase or lowercase letter"
+    )
+    def test_underscore_names(self):
+        pass
```
IlyaFaer, line 152: It tests if the dialect supports table names starting with `_`

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
larkee: Just tested:
```
>>> op = database.update_ddl(["CREATE TABLE {} (col INT64) PRIMARY KEY(col)".format('a'*129)])
>>> op.result()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/pottinger/.pyenv/versions/3.8-dev/lib/python3.8/site-packages/google/api_core/future/polling.py", line 134, in result
    raise self._exception
google.api_core.exceptions.InvalidArgument: 400 Table name not valid: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.
>>> op = database.update_ddl(["CREATE TABLE {} (col INT64) PRIMARY KEY(col)".format('a'*128)])
>>> op.result()

>>> 
```
So it seems the range is inclusive of 128 i.e. the max length is 128. I'll file an issue with the docs to have them be more explicit.

IlyaFaer: Okay, pushed

larkee (approved):

PR #22, opened 2021-03-11
test: add DDL tests group into the test suite

IlyaFaer:


IlyaFaer:
All the tests of DDL testing group are processed - passing fine or skipped.

Skips reasons:
Named schemas and table comments are not supported by Cloud Spanner. Also, max id length in Spanner is 128, which means all the `LongName` tests are not passing - ids are truncated by Spanner to 128 symbols, so they don't match the ones expected by the tests.

Безымянный

larkee (approved):
LGTM :+1:  Just out of curiosity, how are `test_add_table_commit`, `test_create_table_schema` and `test_drop_table_comment` are being skipped? They aren't explicitly skipped in this PR

IlyaFaer:
@larkee, about half of all compliance tests were skipped by default - they require a property to be set in `Requirements()` class. For example, to unskip `test_create_table_schema` (and several more tests, which are using named schemas), one must add one of those exclusions for `schemas` feature:

```python
@property
def schemas(self):
    return exclusions.open()
```
That property will explicitly say that the feature is supported by the dialect and need to be tested. Looking at the tests skipped by default, I'd say only basic features are tested by default. All that is not very-very ordinary requires an explicit enabling.

PR #23, opened 2021-03-11
test: fix string compliance tests

HemangChothani:


test/test_suite.py
```diff
@@ -198,3 +199,81 @@ class TableDDLTest(_TableDDLTest):
...
+                if filter_ is not None:
+                    value = filter_(value)
+                assert value in output
+
+    def test_literal_backslashes(self):
+        """
+        SPANNER OVERRIDE:
+
+        Spanner supports `\\` backslash to represent valid escape sequence,
+        for `'\'` spanner throws an error `400 Syntax error: Illegal escape sequence`.
```
larkee, line 256:
```suggestion
        for `'\'` Spanner throws an error `400 Syntax error: Illegal escape sequence`.
```

test/test_suite.py
```diff
@@ -198,3 +199,81 @@ class TableDDLTest(_TableDDLTest):
...
+                value = row[0]
+                if filter_ is not None:
+                    value = filter_(value)
+                assert value in output
+
+    def test_literal_backslashes(self):
+        """
+        SPANNER OVERRIDE:
+
+        Spanner supports `\\` backslash to represent valid escape sequence,
```
larkee, line 255:
```suggestion
        Spanner supports `\\` backslash to represent valid escape sequence, but
```

test/test_suite.py
```diff
@@ -198,3 +199,81 @@ class TableDDLTest(_TableDDLTest):
...
+        for `'\'` spanner throws an error `400 Syntax error: Illegal escape sequence`.
+        see: https://cloud.google.com/spanner/docs/lexical#string_and_bytes_literals
```
larkee, line 257:
```suggestion
        See: https://cloud.google.com/spanner/docs/lexical#string_and_bytes_literals
```

test/test_suite.py
```diff
@@ -198,3 +199,81 @@ class TableDDLTest(_TableDDLTest):
...
+        self._literal_round_trip(String(40), [data], [data1])
+
+    def test_literal_quoting(self):
+        """
+        SPANNER OVERRIDE:
+
+        The original test string is : \"""some 'text' hey "hi there" that's text\"""
+
+        Spanner doesn't support string which contains `'s` in
+        sentence and throws an of syntax error. e.g.: `'''hey that's text'''`
```
larkee, line 271:
```suggestion
        strings and throws a syntax error, e.g.: `'''hey that's text'''`
```

larkee (commented):
The literal string overrides LGTM, just some nits on the docstrings.

The cleanup override I'm not sure about as mentioned in #20. Let's resolve the discussion there and then apply any decision we make to this PR :+1:

test/test_suite.py
```diff
@@ -198,3 +199,81 @@ class TableDDLTest(_TableDDLTest):
...
+        """
+        SPANNER OVERRIDE:
+
+        Spanner is not able cleanup data and drop the table correctly,
+        table was already exists after related tests finished, so it doesn't
+        create a new table and when started new tests, following
+        insertions will fail with `400 Duplicate name in schema: t.
+        Overriding the tests to create a new table for test and drop table manually
+        before it creates a new table to avoid the same failures.
+        """
```
larkee, line 216: minor nit:
```
        SPANNER OVERRIDE:
        Spanner is not cleaning up data and dropping the table correctly,
        as the DDL statements are being queued, so it doesn't
        create a new table and when started new tests, following
        insertions will fail with `400 Duplicate name in schema: t.
        Overriding the tests to create a new table for test and drop table manually
        before it creates a new table to avoid the same failures.
```

Please add a TODO to remove this override as mentioned in the other PR.

larkee: Disregard

larkee:
The `test_literal_backslashes` test is failing with:
```
sqlalchemy.exc.IntegrityError: (google.cloud.spanner_dbapi.exceptions.IntegrityError) 400 Duplicate name in schema: t_string.
```

test/test_suite.py
```diff
@@ -509,3 +510,85 @@ def _literal_round_trip(self, type_, input_, output, filter_=None):
...
+            for value in input_:
+                ins = (
+                    t.insert()
+                    .values(x=literal(value))
+                    .compile(
+                        dialect=db.dialect, compile_kwargs=dict(literal_binds=True),
+                    )
+                )
+                conn.execute(ins)
+                conn.execute("SELECT 1")
```
larkee, line 550: Insert statements are DML, not DDL so this select isn't needed.
```suggestion
```

I noticed this was also in the integer compliance PR after I merged it (whoops!) so it should probably be removed from there too.

larkee (approved):

HemangChothani:
The `test_literal_backslashes` test will pass after merging https://github.com/cloudspannerecosystem/python-spanner-sqlalchemy/pull/34/ PR

larkee:
> The `test_literal_backslashes` test will pass after merging #34 PR

The Github Action says it's already passing against the emulator 👍

test/test_suite.py
```diff
@@ -509,3 +508,83 @@ def _literal_round_trip(self, type_, input_, output, filter_=None):
...
+
+        self._literal_round_trip(String(40), [data], [data1])
+
+    def test_literal_quoting(self):
+        """
+        SPANNER OVERRIDE:
+
+        The original test string is : \"""some 'text' hey "hi there" that's text\"""
+
+        Spanner doesn't support string which contains `'s` in
```
larkee, line 581: This is incorrect. Spanner does support strings containing `'s': https://cloud.google.com/spanner/docs/lexical#literals

larkee: Using the original test string with the Python client:
Input: `"""some 'text' hey "hi there" that's text"""`
Output: `'some \'text\' hey "hi there" that\'s text'`

test/test_suite.py
```diff
@@ -509,3 +508,83 @@ def _literal_round_trip(self, type_, input_, output, filter_=None):
...
+    def test_literal_backslashes(self):
+        """
+        SPANNER OVERRIDE:
+
+        Spanner supports `\\` backslash to represent valid escape sequence, but
+        for `'\'` Spanner throws an error `400 Syntax error: Illegal escape sequence`.
+        See: https://cloud.google.com/spanner/docs/lexical#string_and_bytes_literals
+        """
+        data = r"backslash one \\ backslash \\\\ two end"
+        data1 = r"backslash one \ backslash \\ two end"
```
larkee, line 571: Please be more descriptive with the variable names. Which of these is the input?

larkee (changes requested):

test/test_suite.py
HemangChothani: Yes, python successfully convert that string but spanner throws an error like `400 Syntax error: Expected \")\" or \",\" but got string literal \'text\' [at 1:41]\nINSERT INTO t_string (x) VALUES (\'some \'\'text\'\' hey \"hi there\" that\'\'s text\')\n`. 
I have tried with DBAPI  with the direct string `'some \'text\' hey "hi there" that\'s text'` and it throws the same error.

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -165,9 +167,13 @@ def visit_DECIMAL(self, type_, **kw):
         return "NUMERIC"
 
     def visit_VARCHAR(self, type_, **kw):
+        if type_.length is None:
+            type_.length = 40
```
HemangChothani, line 171: Added default value of String column, when value is `None`, spanner throws an error.

larkee: Having a default length of 40 seems rather arbitrary. Where did you get this number?

HemangChothani: All other tests having that number so for temporary put the 40  and mentioned for the suggestion.

larkee: Please use `MAX' as the default length

HemangChothani:
@googlebot I fixed it.

HemangChothani:
@googlebot I fixed it.

AVaksman:
@googlebot I fixed it.

test/test_suite.py
vi3k6i5: Spanner prod is working with these 3 strings and adding escape characters to them.
$ Report(name=r"1. backslash one \\ backslash \\\\ two end").save()
$ Report(name=r"2. backslash one \ backslash \\ two end").save()
$ Report(name="""3. some 'text' hey "hi there" that's text""").save()
$ Report.objects.all()[0].name
'1. backslash one \\\\ backslash \\\\\\\\ two end'
$ Report.objects.all()[1].name
'2. backslash one \\ backslash \\\\ two end'
$ Report.objects.all()[2].name
'3. some \'text\' hey "hi there" that\'s text'

HemangChothani: I mean when it comes through the DB API it throws an error:

```python
from google.cloud.spanner_dbapi import connect

connection = connect('instance-id', 'databse-id')
cursor = connection.cursor()

cursor.execute("INSERT INTO t (x) VALUES ('backslash one \ backslash \\ two end')")   #it will through an error `400 Syntax error: Illegal escape sequence`

cursor.execute("""INSERT INTO t (x) VALUES (some 'text' hey "hi there" that's text)""")  #it will through an error `400 Syntax error: Expected \")\" or \",\" but got string literal \'text\' `
```

@vi3k6i5 if you are using spanner client directly for the insert, then for help can you share the actual sql query syntax passing at that time.

larkee: Okay, it looks like the issue is the SQL query rather than the data string. Spanner supports both `'backslash one \ backslash \\ two end''` and `"some 'text' hey "hi there" that's text'` however, when specifying these types of strings by SQL, the additional escapes are necessary.

Rather than skipping these tests, we need to add additional logic to add these additional escapes before they are used to generate SQL statements.

larkee: Here is the DBAPI code showing the strings are supported:
```
connection = connect('test-instance', 'test-db')
cursor = connection.cursor()

cursor.execute(
    "INSERT INTO t (pkey, val) VALUES (0, \"{}\")".format(
        r'backslash one \\ backslash two \\\\ end'))

cursor.execute(
    "INSERT INTO t (pkey, val) VALUES (1, \"{}\")".format(
        r"""some 'text' hey \"hi there\" that's text"""))

cursor.execute("SELECT * FROM T")
for row in cursor.fetchall():
    print(str(row[1]))
```
Output:
```
some 'text' hey "hi there" that's text
backslash one \ backslash two \\ end
```

HemangChothani:
@larkee Can you change the `cla` for this PR? Don't know how to changed from yes to no.

test/test_suite.py
```diff
@@ -843,3 +844,38 @@ def test_nolength_binary(self):
 @pytest.mark.skip("Spanner doesn't support quotes in table names.")
 class QuotedNameArgumentTest(_QuotedNameArgumentTest):
     pass
+
+
+class StringTest(_StringTest):
+    def test_literal_backslashes(self):
```
larkee, line 850: With the added `render_literal_value` method, `test_literal_backslashes` and `test_literal_quoting` should no longer need to be overridden. Please remove.

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -137,6 +137,32 @@ def visit_empty_set_expr(self, type_):
...
+        This should be implemented by subclasses using the quoting services
+        of the DBAPI.
+
+        Cloud spanner supports prefixed backslash to escape non-alphanumeric characters
+        in string. Override the method to add  additional escape before using it to
+        generate a SQL statement.
+        """
+        raw = ["\\", "'", '"', "\n", "\t", "\r"]
+        if type(value) == str and any(single in value for single in raw):
+            value = '"{}"'.format(value)
```
larkee, line 155: Locally, after removed the `test_literal_backslashes` and `test_literal_quoting` overrides, I was still getting failures.
The following change fixes the errors:
```suggestion
            value = 'r"""{}"""'.format(value)
```

larkee (approved):

PR #26, opened 2021-03-16
fix: run compliance tests with nox

### LINKED ISSUE #25: compliance tests are failing
Compliance tests are not running with new version of `sqlalchemy`
## Expected Behavior
Successfully run the tests

## Actual Behavior
Throw as error `sqlalchemy.exc.ArgumentError: Could not parse rfc1738 URL from string 'None'`

## Steps to Reproduce the Problem

1. nox -s compliance_test

## Specifications

- Version: sqlalchemy== 1.4.0

HemangChothani:
Fixes #25

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ setuptools.setup(
     description=description,
     entry_points={
         "sqlalchemy.dialects": [
-            "spanner = google.cloud.sqlalchemy_spanner:SpannerDialect"
+            "spanner.spanner = google.cloud.sqlalchemy_spanner:SpannerDialect"
         ]
     },
     install_requires=dependencies,


setup.py
```diff
@@ -41,7 +41,7 @@
     description=description,
     entry_points={
         "sqlalchemy.dialects": [
-            "spanner = google.cloud.sqlalchemy_spanner:SpannerDialect"
+            "spanner.spanner = google.cloud.sqlalchemy_spanner:SpannerDialect"
```
IlyaFaer, line 44: It'll break the package. This identifier is used to register dialect in Python modules system. With this change we'll have to connect like:
```python
engine = create_engine(
    "spanner.spanner:///projects/project-id/instances/instance-id/databases/database-id"
)
```
And I don't think it'll work at all.

HemangChothani: I think it's working fine as you can see in SQLAlchemy spanner dialect test.

The url is `spanner+spanner:///projects/appdev-soda-spanner-staging/instances/sqlalchemy-dialect-test/databases/compliance-test`, which is perfectly working now.

`spanner.spanner` is used only to fetch the Dialect class. New version of `sqlalchemy` used by default `spanner+spanner` and replace `+ ` to `.` to fetch the entry-point or dialect class.

HemangChothani: The other way to overcome the failure is to override the method which added or forcefully converted `spanner` to `spanner+spanner`.

The method is `generate_driver_url`  and path is `sqlalchemy/testing/provision.py`

IlyaFaer: I've checked the tests locally, there is no failure at all.

HemangChothani: For now compliance tests are failing in you latest opened PR
https://github.com/cloudspannerecosystem/python-spanner-sqlalchemy/pull/27/checks?check_run_id=2119935389 

Cloud you help me to solve that issue?

IlyaFaer: Here is the local run - everything is working fine on the last SQLAlchemy version without these or any other changes. The most meaningful thing to fix right now is the insert failure of PR 223. Without it we're not able to use GitHub workflow in any case.

Безымянный

HemangChothani: I know that's working fine, but when run `nox -s compliance_test` locally, it's unable get the entry point and unable to create a url for engine and throw an error, so can't connect to  the server.  
 Screenshot from 2021-03-16 15-30-04

IlyaFaer: That means the problem is in nox environment. Most likely in package setup part, not in the package registering.

HemangChothani: I think the problem i have explained before in the comment https://github.com/cloudspannerecosystem/python-spanner-sqlalchemy/pull/26#discussion_r594994290

Better to pin older version of `sqlalchemy` if every one will agree.

larkee: Currently I am getting the same error as @HemangChothani which is also showing up on the Github Action. I'm not sure why it didn't show up before given that it does seem to be a issue with `noxfile.py`.

Locally, pinning to an older version of `sqlalchemy` is not working on my end.

@IlyaFaer How are you running the compliance tests locally? Are you using the nox command? If not, please post the commands that you are running here so we can update the noxfile to match.

HemangChothani: @larkee  In setup.py file I replace `dependencies = ["sqlalchemy>=1.1.13"]` to dependencies = ["sqlalchemy>=1.1.13, <=1.3.23"] and it works for me.

The reason of failing the command is they have implemented new method  `generate_driver_url()` and path is `sqlalchemy/testing/provision.py` to generate the url for engine. In that method they replace spanner to spanner+spanner and going to find entry point of the library, in older version there was a spanner as a drive-name so they found perfectly, but in the new version they replace `+` with `.` and going to find entry point with `spanner.spanner` but we have `spanner` in setup file so it returns `None` that's why it's failing.

IlyaFaer: @larkee, I'm executing tests locally just as usual, the only difference - I don't install packages on every nox session run, I'm using the local ones:

Безымянный

But if I'll uncomment these installing lines in nox session function, everything will still work.

As a simple select-script (executed without nox) is working fine with any version of SQLAlchemy, we can say registering the dialect should not to be changed (that would be strange, if SQLAlchemy team would change dialect registering mechanism in 1.4 version - it would break all the dialects). But nox session is failing, because registering mechanism is not able to find the Spanner dialect class. That makes me think the problem is in installing spanner-sqlalchemy package by nox.

HemangChothani: @IlyaFaer  I have tried with manually installed `spanner-sqlalchemy` and `SQLAlchemy`  and run  pytest command manually  (WIthout nox) and still `sqlalchemy` raising the same error. I think they have changed few things in testing procedure.

larkee: I can confirm that updating `setup.py` with: 
```
dependencies = ["sqlalchemy>=1.1.13, <=1.3.23", "google-cloud-spanner>=3.0.0"]
```
fixes the issue. The update to entry point is unneeded.

Two additional notes:
1) The `compliance_test` session does not need to explicitly install `sqlalchemy` or `google-cloud-spanner` because they are automatically install by `session.install("-e", ".")` which installs `spanner-sqlalchemy` using `setup.py`
2) `create_test_database` uses the project set by either `GOOGLE_CLOUD_PROJECT` or `PROJECT_ID` or defaults to `emulator-test-project` but the connection string defined in `setup.cfg` using the prod project.

larkee (commented):
The issues definitely seems to be sqlalchemy v1.4.0 since it was released yesterday and that's when this issue started appearing so I'm happy for the version to be pinned to `sqlalchemy>=1.1.13, <=1.3.23` with no other changes needed :+1:

HemangChothani:
Closing this PR in favor of https://github.com/cloudspannerecosystem/python-spanner-sqlalchemy/pull/31

PR #27, opened 2021-03-16
feat: implement schemas reflection method

IlyaFaer:


google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -310,7 +310,9 @@ def get_indexes(self, connection, table_name, schema=None, **kw):
 FROM information_schema.indexes as i
 JOIN information_schema.index_columns AS ic
     ON ic.index_name = i.index_name AND ic.table_name = i.table_name
-WHERE i.table_name="{table_name}"
+WHERE
+    i.table_name="{table_name}"
+    AND i.index_type != 'PRIMARY_KEY'
```
IlyaFaer, line 315: Locally I've implemented drop table mechanism which first drops indexes. The problem is that it also tries to drop primary keys, which is not correct. Fixing it.

test/test_suite.py
```diff
@@ -42,6 +42,8 @@
 from sqlalchemy.testing.suite.test_select import ExistsTest as _ExistsTest
 from sqlalchemy.testing.suite.test_types import BooleanTest as _BooleanTest
 
+config.test_schema = ""
+
```
IlyaFaer, line 46: That's a property to be set by dialect testing system. As Spanner doesn't support named schemas, using an empty str - the default scheme to run tests.

IlyaFaer:
These changes are making two more reflection tests to pass. The first one was passing, but its cleanup was failing, making other tests to fail.

Безымянный

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
larkee: This fixes the drop table mechanism that you've implemented. Is this method used elsewhere? Is this a method that users will call? If so, could you confirm that this method should not be returning primary key indexes?

IlyaFaer: Well, it's not directly related to the drop table mechanism, the method is for introspection. Still, it's used by a bunch of methods, including the one dropping tables.

Users can use this method - they can instant an `Inspector()` object and use it for introspection. As primary key indexes are something controlled by Spanner itself, I assume we can avoid showing them to user, considering it as an obvious and auto regulated thing, which works without user's control.

larkee: If other database products such as MySQL do not return primary key indexes then this LGTM :+1:  Please confirm that this is the case

IlyaFaer: Yeah, here is a couple of examples

SQLite:
https://github.com/sqlalchemy/sqlalchemy/blob/76b33bb435f1f55111cd68cde9b7cb735c748bcd/lib/sqlalchemy/dialects/sqlite/base.py#L2454

MySQL:
https://github.com/sqlalchemy/sqlalchemy/blob/76b33bb435f1f55111cd68cde9b7cb735c748bcd/lib/sqlalchemy/dialects/mysql/base.py#L3170-L3171

PostgreSQL:
https://github.com/sqlalchemy/sqlalchemy/blob/3fec5028e695ad138aa46a0ae66c55e8bcb653f6/lib/sqlalchemy/dialects/postgresql/base.py#L4048

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -372,6 +374,33 @@ def get_pk_constraint(self, connection, table_name, schema=None, **kw):
...
+
+        Args:
+            connection (Union[
+                sqlalchemy.engine.base.Connection,
+                sqlalchemy.engine.Engine
+            ]):
+                SQLAlchemy connection or engine object.
+
+        Returns:
+            list: Schemas' names.
```
larkee, line 388:
```suggestion
            list: Schema names.
```

larkee (approved):

PR #29, opened 2021-03-16
test: fix numeric compliance test

HemangChothani:


larkee (commented):
There's a lot of skipping relating to decimal which I'm concerned about. Would the changes included in #33 be relevant here?

HemangChothani:
All the tests passes once https://github.com/googleapis/python-spanner/pull/290 PR merged.

vi3k6i5:
Tests look good to me. LGTM.

vi3k6i5 (approved):
Test changes look good to me. Please wait for Skylar's approval before merging.

larkee (commented):
LGTM other than the nit

google/cloud/sqlalchemy_spanner/requirements.py
```diff
@@ -68,3 +68,25 @@ def get_isolation_levels(self, _):
...
+    @property
+    def precision_numerics_many_significant_digits(self):
+        """target backend supports values with many digits on both sides,
+        such as 319438950232418390.273596, 87673.594069654243
+
+        """
+        return exclusions.open()
+
+    @property
+    def precision_numerics_retains_significant_digits(self):
```
larkee, line 81: Based on this override reason:
```
SPANNER OVERRIDE:
        Cloud Spanner ignores the `0`s after the decimal point if all
        the digits are `0`.
        Overriding the test to avoid the assertion failure.
```

We should not have this property open as it enables a single test that we have to override.

google/cloud/sqlalchemy_spanner/requirements.py
```diff
@@ -68,3 +68,25 @@ def get_isolation_levels(self, _):
             dict: isolation levels description.
         """
         return {"default": "SERIALIZABLE", "supported": ["SERIALIZABLE", "AUTOCOMMIT"]}
+
+    @property
+    def precision_numerics_many_significant_digits(self):
```
larkee, line 73: This property is not true for Spanner so we should not be setting it for open as the test will fail without overrides.

HemangChothani: This is supported, but till scale =  9.

larkee: Right, but one of the example numbers is scale = 12. I worry that having this open suggests that we support arbitrary precision `NUMERIC` which is not true.

From a quick search I found [this](https://github.com/zzzeek/sqlalchemy/blob/master/test/requirements.py#L1213) so I still think we should close this property.

test/test_suite.py
```diff
@@ -850,3 +851,320 @@ class StringTest(_StringTest):
...
+        only a single row can be inserted into such a table -
+        following insertions will fail with `Row [] already exists".
+        Overriding the test to avoid the same failure.
+
+        Remove an extra digits after decimal point as cloud spanner is
+        capable of representing an exact numeric value with a precision
+        of 38 and scale of 9.
+        """
+        self._do_test(
+            Numeric(precision=18, scale=12),
```
larkee, line 1029: We are overriding a number of tests because Spanner does not support `scale>9`. But these tests still uses `Numeric` with `scale=12` or `scale=14`. I think it makes sense to change these `Numeric`s to use `scale=9`. WDYT?

larkee (approved):

PR #32, opened 2021-03-19
fix: use empty schema name by default

IlyaFaer:
The PR is based on another: https://github.com/cloudspannerecosystem/python-spanner-sqlalchemy/pull/24

PR #33, opened 2021-03-19
fix: types reflection

IlyaFaer:


google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -33,7 +33,7 @@
     "DATETIME": types.DATETIME,
     "FLOAT64": types.Float,
     "INT64": types.BIGINT,
-    "NUMERIC": types.DECIMAL,
+    "NUMERIC": types.NUMERIC(precision=38, scale=9),
```
IlyaFaer, line 36: Constant precision and scale

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -164,6 +176,9 @@ def visit_BINARY(self, type_, **kw):
     def visit_DECIMAL(self, type_, **kw):
         return "NUMERIC"
 
+    def visit_NUMERIC(self, type_, **kw):
+        return "NUMERIC"
```
IlyaFaer, line 180: With this change NUMERIC type will be created without precision and scale mentions. Even if user will set something with explicit dimensions, for example, `types.Numeric(precision=2, scale=2)`, in DDL only keyword `NUMERIC` will be used. Otherwise syntax error will fail.

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -302,12 +317,17 @@ def get_columns(self, connection, table_name, schema=None, **kw):
             columns = snap.execute_sql(sql)
 
             for col in columns:
-                type_ = "STRING" if col[1].startswith("STRING") else col[1]
+                if col[1].startswith("STRING"):
+                    end = col[1].index(")")
+                    size = int(col[1][7:end])
+                    type_ = _type_map["STRING"](length=size)
+                else:
+                    type_ = _type_map[col[1]]
```
IlyaFaer, line 325: While reflecting column types, we're reading strings like `STRING(size)". Such a string can be parsed to get the columns size - to set it into SQLAlchemy reflected type.

test/test_suite.py
```diff
@@ -509,3 +550,196 @@ def _literal_round_trip(self, type_, input_, output, filter_=None):
...
+
+        Spanner defines NUMERIC type with the constant precision=38
+        and scale=9. Overriding the test to check if the NUMERIC
+        column is successfully created and has dimensions
+        correct for Cloud Spanner.
+        """
+        for typ in self._type_round_trip(Numeric(18, 5)):
+            assert isinstance(typ, Numeric)
+            eq_(typ.precision, 38)
+            eq_(typ.scale, 9)
```
IlyaFaer, line 745: In this override I'm keeping the original `(18, 5)` dimensions to check that dialect will create NUMERIC without trying to do it like `NUMERIC(18, 5)`, as it'll cause syntax error in Spanner.

IlyaFaer:
@larkee, as this PR is based on #32 and 32 is based on #24, the branch includes all the changes of my last three PRs. With all of this I have only three reflections tests failing. Most of the skipped tests are related to the named schemas feature, so they are going to stay skipped. Thus, we're almost there with reflection I suppose.

image

IlyaFaer:
@larkee, here I'm also adding `results` testing group into the test suite, as all the tests of the group seem to be processed:

`PercentSchemaNamesTest` are test cases to test if `%` symbol can be used in column, table, schema names. `%` can't be used in Spanner identifier, so leaving the tests skipped.

`ServerSideCursorsTest` - I didn't find any mentions of server cursors in Spanner, so I assume it's not supported, and the test case can stay skipped.

Безымянный

AVaksman (approved):

PR #35, opened 2021-03-25
Update README.md

IlyaFaer:


larkee (changes requested):
Please add information regarding DDL to the limitation section. Specifically that they are not executed as part of the transaction and as such cannot be rolled back.

README.md
```diff
@@ -15,3 +44,66 @@ table = Table("table-id", MetaData(bind=engine), autoload=True)
...
+    metadata,
+    Column('col1', Integer),
+    Index("uix_1", "col1", unique=True),
+)
+```
+**Autocommit mode**  
+Spanner dialect supports both "autocommit" and "manual commit" modes. To set/change the current mode isolation levels can be used: `SERIALIZABLE` and `AUTOCOMMIT`.
+
+**DDL and transactions**  
+DDL statements are executed aside of regular transactions mechanism, which means DDL statements will not be rolled back on normal transaction rollback.
```
skuruppu, line 100:
```suggestion
DDL statements are executed outside the regular transactions mechanism, which means DDL statements will not be rolled back on normal transaction rollback.
```

README.md
```diff
@@ -15,3 +44,66 @@ table = Table("table-id", MetaData(bind=engine), autoload=True)
...
+
+**Dropping a table**  
+Cloud Spanner, by default, doesn't drop tables, which has secondary indexes and/or foreign key constraints. In Spanner dialect for SQLAlchemy, however, this restriction is omitted - if a table you are trying to delete has indexes/foreign keys, they will be dropped automatically right before dropping the table.
+
+**Other limitations**  
+- WITH RECURSIVE statement is not supported.
+- Named schemas are not supported.
+- Temporary tables are not supported, real tables are used instead.
+- Numeric type dimensions (scale and precision) are constant. See the [docs](https://cloud.google.com/spanner/docs/data-types#numeric_types).
```
skuruppu, line 109: Some other sections to include either here or in a separate file:
- Type mapping: I think people would like to have a reference for this in the README instead of having to look through the code.
- Information or any pointers on how to migrate an existing database into CSpanner.
- Running the tests against prod and/or the emulator: this can also go in CONTRIBUTING.md.
- License link.

skuruppu (changes requested):

README.md
IlyaFaer: Done

README.md
```diff
@@ -15,3 +44,101 @@ table = Table("table-id", MetaData(bind=engine), autoload=True)
...
+    metadata,
+    Column("user_id", Integer, primary_key=True),
+    Column("user_name", String(16), nullable=False),
+)
+
+metadata.create_all(engine)
+```
+Migration
+-----------
+SQLAlchemy uses [Alembic](https://alembic.sqlalchemy.org/en/latest/#) tool to organize database migrations.
```
skuruppu, line 75: Is this going to work out of the box or is there manual adjustments a user would need to make to get it working with Cloud Spanner?

skuruppu (approved):

README.md
IlyaFaer: It works out of the box, as Alembic is built on top of SQLAlchemy, and inside a migration script simple SQLAlchemy commands are used. Meaning, if SQLAlchemy works fine with Spanner than Alembic should also work fine. I assume we should deal with the last data types problems in the dialect, and then we'll be able to write tests for migrations.

README.md
```diff
@@ -15,3 +44,101 @@ table = Table("table-id", MetaData(bind=engine), autoload=True)
...
+)
+```
+**Autocommit mode**  
+Spanner dialect supports both "autocommit" and "manual commit" modes. To set/change the current mode isolation levels can be used: `SERIALIZABLE` and `AUTOCOMMIT`.
+
+**DDL and transactions**  
+DDL statements are executed outside the regular transactions mechanism, which means DDL statements will not be rolled back on normal transaction rollback.
+
+**Dropping a table**  
+Cloud Spanner, by default, doesn't drop tables, which has secondary indexes and/or foreign key constraints. In Spanner dialect for SQLAlchemy, however, this restriction is omitted - if a table you are trying to delete has indexes/foreign keys, they will be dropped automatically right before dropping the table.
```
larkee, line 107:
```suggestion
Cloud Spanner, by default, doesn't drop tables, which have secondary indexes and/or foreign key constraints. In Spanner dialect for SQLAlchemy, however, this restriction is omitted - if a table you are trying to delete has indexes/foreign keys, they will be dropped automatically right before dropping the table.
```

larkee (approved):

PR #39, opened 2021-03-30
test: add basic circleci config

skuruppu:


.circleci/config.yml
```diff
@@ -0,0 +1,21 @@
+version: 2.1
+
+orbs:
+  python: circleci/python@0.2.1
```
IlyaFaer, line 4: I'm not very familiar with this kind of ci yet, but seems like everything is okay here. The only thing I've noticed is version: looks like the 0.2.1 is old a little bit, and the current version is [1.3.3](https://circleci.com/developer/orbs/orb/circleci/python)

IlyaFaer (approved):

.circleci/config.yml
skuruppu: I'll try changing it here. I suspect we'll have to make lots of modifications to the default config to get it running anyway.

PR #41, opened 2021-03-30
test: unicode compliance tests

HemangChothani:


test/test_suite.py
```diff
@@ -549,6 +551,98 @@ def _literal_round_trip(self, type_, input_, output, filter_=None):
...
+            unicode_table.insert(),
+            [{"id": i, "unicode_data": self.data} for i in range(3)],
+        )
+
+        rows = config.db.execute(select([unicode_table.c.unicode_data])).fetchall()
+        eq_(rows, [(self.data,) for i in range(3)])
+        for row in rows:
+            assert isinstance(row[0], util.text_type)
+
+    @pytest.mark.skip("Spanner throws an error")
```
larkee, line 596: I don't think this is a valid reason to skip a test. If Spanner is throwing an error when `sqlalchemy` is not expecting an error then we need to add error handling.

larkee (approved):

PR #43, opened 2021-03-31
fix: update circleci config to run tests

skuruppu:


skuruppu:
@IlyaFaer please take this PR and modify it to get it working with the tests. I'll set up the env vars and report back on what they are.

skuruppu:
Credentials in GOOGLE_APPLICATION_CREDENTIALS variable and project is in GOOGLE_CLOUD_PROJECT.

IlyaFaer:
@skuruppu, okay, it looks like nox sessions are successfully executed, but credentials file is not found. As Spanner dialect is built on top of Spanner DB API and DB API is build on top of Spanner default Python client, it uses auth system of the original Python client. Meaning it should work on the same env vars python-spanner uses.

skuruppu:
Oh interesting. The issue is that you can't store files in CircleCI so you have to somehow dump the key into a file at runtime. I tried with a few commands (which work for me locally) but somehow CircleCI doesn't like it. Probably that it doesn't have the json package installed? I don't know.

I'll have to give up for today and look into it again tomorrow.

@larkee if there's a way to read credentials in the Python client from an env variable, please let us know.

larkee:
The credentials are currently being passed by the `GOOGLE_APPLICATION_CREDENTIALS` env variable. Unfortunately, the file that it points to doesn't exist. However, there is another way for the client to get credentials that does not rely on a file:
https://github.com/googleapis/google-auth-library-python/blob/master/google/auth/_default.py#L368-L379

If the credential file is not specified, the client will try to get credentials via the Google Cloud SDK. I found [this circleci orb](https://circleci.com/developer/orbs/orb/circleci/gcp-cli) which may help.

IlyaFaer:
@larkee, @skuruppu, I've pushed some changes, but that's not enough. As I see, a *context* should be used with the ci config, and I'm not able to open the contexts page (it says that I don't have permissions).

skuruppu:
Hooray 🎉  I think I got it running. There was something weird about the env vars but it's fixed now.

We don't need gcp-cli.

skuruppu:
This is the list of failures now from actually running the tests:

https://app.circleci.com/pipelines/github/cloudspannerecosystem/python-spanner-sqlalchemy/101/workflows/49e96c8e-e45d-4700-bda6-5151e7ffc2d4/jobs/109/parallel-runs/0/steps/0-105

skuruppu:
@IlyaFaer please add some code in the tests to either detect that the test instance exists or to delete the old instance before attempting to create a new one. Otherwise, every subsequent run, you get this error:

`google.api_core.exceptions.AlreadyExists: 409 Instance already exists: projects/***************************/instances/sqlalchemy-dialect-test`

https://app.circleci.com/pipelines/github/cloudspannerecosystem/python-spanner-sqlalchemy/102/workflows/9eb61658-ec9c-4c17-8d0a-9e1884f0c133/jobs/111/parallel-runs/0/steps/0-105

AVaksman (approved):

PR #46, opened 2021-04-05
test: skip unsupported features tests

IlyaFaer:
Spanner doesn't support temporary tables. The same tests (for indexes, table names, etc.) are passing on real tables, so these tests can be skipped.

### DIFF
--- a/google/cloud/sqlalchemy_spanner/requirements.py
+++ b/google/cloud/sqlalchemy_spanner/requirements.py
@@ -38,6 +38,10 @@ class Requirements(SuiteRequirements):
         return exclusions.open()
 
     @property
+    def implements_get_lastrowid(self):
+        return exclusions.closed()
+
+    @property
     def ctes(self):
         return exclusions.open()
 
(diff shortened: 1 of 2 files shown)


test/test_suite.py
```diff
@@ -682,3 +685,17 @@ def test_unique_constraint_raises(self):
...
+
+    @pytest.mark.skip("Spanner doesn't support temporary tables")
+    def test_get_temp_table_unique_constraints(self):
+        pass
+
+
+class InsertBehaviorTest(_InsertBehaviorTest):
+    @pytest.mark.skip("Spanner doesn't support empty inserts")
+    def test_empty_insert(self):
+        pass
```
IlyaFaer, line 701: Also skippting a test with an empty insert. See [the docs](https://cloud.google.com/spanner/docs/dml-syntax#insert-statement).

google/cloud/sqlalchemy_spanner/requirements.py
```diff
@@ -37,6 +37,10 @@ def autocommit(self):
     def order_by_collation(self):
         return exclusions.open()
 
+    @property
+    def implements_get_lastrowid(self):
+        return exclusions.closed()
```
IlyaFaer, line 42: `lastrowid` is not supported

AVaksman (approved):

PR #47, opened 2021-04-06
test: select compliance tests

HemangChothani:


AVaksman (approved):

test/test_suite.py
```diff
@@ -837,3 +835,91 @@ def test_nolength_binary(self):
...
+
+
+@pytest.mark.skip("Spanner doesn't support IS DISTINCT FROM clause")
+class IsOrIsNotDistinctFromTest(_IsOrIsNotDistinctFromTest):
+    pass
+
+
+class OrderByLabelTest(_OrderByLabelTest):
+    @pytest.mark.skip(
+        "Spanner throws a syntax error as group by composed doesn't support."
```
larkee, line 847: I'm not quite sure what isn't supported but perhaps it should be reworded to:
```Spanner doesn't support <unsupported statement>.```

Similar to the above skip

File: test/test_suite.py [hunk 2]
```diff
@@ -837,3 +835,91 @@ def test_nolength_binary(self):
...
+    )
+    def test_group_by_composed(self):
+        pass
+
+
+class CompoundSelectTest(_CompoundSelectTest):
+    @pytest.mark.skip("Spanner throws a syntax error.")
```
larkee, line 854: Please reword these to `Spanner doesn't support <unsupported statement>.`

test/test_suite.py
```diff
@@ -837,3 +835,91 @@ def test_nolength_binary(self):
...
+    def test_limit_offset_selectable_in_unions(self):
+        pass
+
+    @pytest.mark.skip("Spanner throws a syntax error.")
+    def test_order_by_selectable_in_unions(self):
+        pass
+
+
+class LikeFunctionsTest(_LikeFunctionsTest):
```
larkee, line 863: Cloud Spanner does not currently support `ESCAPE`. That's what's causing the "parenthesized expression" error. Please update the skip reason to state Spanner does not support `ESCAPE`.

test/test_suite.py
larkee: Running these locally, I am getting the syntax error however I think the intended SQL statement should be supported. It looks like the argument replacement is not working as intended.

Error:
```
sqlalchemy.exc.ProgrammingError: (google.cloud.spanner_dbapi.exceptions.ProgrammingError) 400 Syntax error: Expected end of input but got keyword WHERE [at 8:12]\n LIMIT @a4 WHERE 1=1\n
[SQL: (SELECT some_table.id, some_table.x, some_table.y 
FROM some_table 
WHERE some_table.id = %s ORDER BY some_table.id
 LIMIT %s) UNION DISTINCT (SELECT some_table.id, some_table.x, some_table.y 
FROM some_table 
WHERE some_table.id = %s ORDER BY some_table.id
 LIMIT %s) ORDER BY id
 LIMIT %s]
[parameters: [2, 1, 3, 1, 2]]
```
Valid SQL query in Spanner:
```
(SELECT some_table.id, some_table.x, some_table.y 
FROM some_table 
WHERE some_table.id = 2 ORDER BY some_table.id
 LIMIT 1) UNION DISTINCT (SELECT some_table.id, some_table.x, some_table.y 
FROM some_table 
WHERE some_table.id = 3 ORDER BY some_table.id
 LIMIT 1) ORDER BY id
 LIMIT 2
```

This may be a bug in DBAPI. Could you look into this?

HemangChothani: sure.

HemangChothani: @larkee Query is not valid in spanner which you have mentioned above, it throws an error due to `GROUP BY` and `LIMIT` in subqueries and  brackets `()`  used for second subqueries.

test/test_suite.py
HemangChothani: Done.

test/test_suite.py
larkee: The query with the placeholders replaced by the parameters is valid, as per Pantheon:
image

However `spanner_dbapi` is throwing an error for the query, which indicates there is a bug in the `spanner_dbapi` logic.

From the exception:
```
400 Syntax error: Expected end of input but got keyword WHERE [at 8:12]\n LIMIT @a4 WHERE 1=1\n
```
It looks like a dummy `WHERE` clause is being added incorrectly.

The statement is being classified as `STMT_UPDATING` rather than `STMT_NON_UPDATING` as the regex expressions currently do not account for SQL that begins with brackets.

larkee: I've opened an issue to track this bug so if you update the skip reason to reference the issue then I'm happy to approve and merge this PR.

The issue: https://github.com/googleapis/python-spanner/issues/347

test/test_suite.py
```diff
@@ -1339,3 +1323,67 @@ def test_enotation_decimal(self):
...
+    pass
+
+
+@pytest.mark.skip("Spanner doesn't support IS DISTINCT FROM clause")
+class IsOrIsNotDistinctFromTest(_IsOrIsNotDistinctFromTest):
+    pass
+
+
+class OrderByLabelTest(_OrderByLabelTest):
+    @pytest.mark.skip("Spanner doesn't supports composed GROUP BY.")
```
larkee, line 1377: This skip reason is not quite correct.

For the given query:
```
SELECT count(some_table.id) AS count_1, some_table.x + some_table.y AS lx 
FROM some_table GROUP BY some_table.x + some_table.y ORDER BY lx
```
Spanner isn't recognizing that the `some_table.x + some_table.y` in the `GROUP BY` list is the same as the `some_table.x + some_table.y` in the `SELECT` list.

The explanation I was given:
> You can use the derived expression in the GROUP BY, but if you want to use it in the SELECT list too, you have to use an alias or some other sort of label to clue Spanner in on the fact that they're the same thing.

So the valid version of this query would be:
```
SELECT count(some_table.id) AS count_1, some_table.x + some_table.y AS lx 
FROM some_table GROUP BY lx ORDER BY lx
```

I think this may be a more accurate message:
`Spanner requires an alias for the GROUP BY list when specifying derived columns also used in SELECT`

larkee (approved):

PR #48, opened 2021-04-08
fix: support types with length

IlyaFaer:


google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -165,13 +165,16 @@ def visit_FLOAT(self, type_, **kw):
         return "FLOAT64"
 
     def visit_TEXT(self, type_, **kw):
-        return "STRING({})".format(type_.length)
+        return "STRING({})".format(type_.length or "MAX")
```
IlyaFaer, line 168: Setting `MAX` value in case if type length wasn't explicitly set

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -165,13 +165,16 @@ def visit_FLOAT(self, type_, **kw):
...
 
     def visit_ARRAY(self, type_, **kw):
         return "ARRAY<{}>".format(self.process(type_.item_type, **kw))
 
     def visit_BINARY(self, type_, **kw):
-        return "BYTES"
+        return "BYTES({})".format(type_.length or "MAX")
+
+    def visit_large_binary(self, type_, **kw):
+        return "BYTES({})".format(type_.length or "MAX")
```
IlyaFaer, line 177: The last SQLAlchemy version says that `Binary` type is deprecated, and `LargeBinary` should be used instead. The `LargeBinary` type will pass `BLOB` into SQL, so I'm overriding the method to make it `BYTES`.

test/test_suite.py
```diff
@@ -790,3 +793,14 @@ class InsertBehaviorTest(_InsertBehaviorTest):
...
+
+class BytesTest(_LiteralRoundTripFixture, fixtures.TestBase):
+    __backend__ = True
+
+    def test_nolength_string(self):
+        metadata = MetaData()
+        foo = Table("foo", metadata, Column("one", LargeBinary))
+
+        foo.create(config.db)
+        foo.drop(config.db)
```
IlyaFaer, line 806: There was no test for binary type without length - adding it

AVaksman (approved):

PR #50, opened 2021-04-13
fix: use INT64 type for empty sets with no type

IlyaFaer:


IlyaFaer:
Tiny fix for empty set expressions: use INT64 in case if type of the empty set wasn't explicitly mentioned. Makes one more test passing:

Безымянный

AVaksman (approved):

PR #51, opened 2021-04-13
feat: add opentelemetry tracing

HemangChothani:


larkee (commented):
Please add tests.

Additionally, this does not not yet add tracing. The `trace_call` needs to be called within the code to create the spans. See [here](https://github.com/googleapis/python-spanner/pull/107) for more context.

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -761,4 +762,31 @@ def do_rollback(self, dbapi_connection):
         ):
             pass
         else:
-            dbapi_connection.rollback()
+            trace_attributes = {"db.connection": dbapi_connection}
```
larkee, line 765: `dbapi_connection` is an object which means the trace attribute is going to show a value like: 
```
"dbapi_connection": <google.cloud.spanner_dbapi.connection.Connection object at 0x7fd1b01e3880>
```

Instead, set the `db.instance` attribute with `dbapi_connection.database.name` so that the span will have the URI of the database, e.g.
```
"db.instance": "projects/<project>/instances/<instance>/databases/<database>"
```

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -761,4 +762,31 @@ def do_rollback(self, dbapi_connection):
...
+        with trace_call("SpannerSqlAlchemy.Commit", trace_attributes):
+            dbapi_connection.commit()
+
+    def do_close(self, dbapi_connection):
+        trace_attributes = {"db.instance": dbapi_connection.database.name}
+        with trace_call("SpannerSqlAlchemy.Close", trace_attributes):
+            dbapi_connection.close()
+
+    def do_executemany(self, cursor, statement, parameters, context=None):
+        trace_attributes = {"db.statement": statement, "db.params": parameters}
```
larkee, line 780: Please also include the `db.instance` attribute here and in the methods below. You should be able to get the database name via `cursor.connection.database.name`.  This makes the span much more useful as it tells the user which database the statement was run against.

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -761,4 +762,42 @@ def do_rollback(self, dbapi_connection):
...
+        with trace_call("SpannerSqlAlchemy.Close", trace_attributes):
+            dbapi_connection.close()
+
+    def do_executemany(self, cursor, statement, parameters, context=None):
+        trace_attributes = {
+            "db.statement": statement,
+            "db.params": parameters,
+            "db.instance": cursor.connection.database.name,
+        }
+        with trace_call("SpannerSqlAlchemy.Executemany", trace_attributes):
```
larkee, line 785: nit:
```suggestion
        with trace_call("SpannerSqlAlchemy.ExecuteMany", trace_attributes):
```

larkee (approved):
LGTM with the suggestion.

I would like some unit tests for this but they can come later. Please create an issue to track this.

HemangChothani:
Issue for tracking : https://github.com/cloudspannerecosystem/python-spanner-sqlalchemy/issues/76

PR #52, opened 2021-04-14
refactor: use a decorator for connection initiation

IlyaFaer:


IlyaFaer:
@larkee, as you proposed once, adding a decorator to deal with `Engine` and initiate a connection.

AVaksman (approved):

PR #54, opened 2021-04-19
test: skip test, which checks use of quotes in table names

IlyaFaer:


test/test_suite.py
```diff
@@ -564,6 +567,11 @@ class SequenceTest(_SequenceTest):
     pass
 
 
+@pytest.mark.skip("Spanner doesn't support quotes in table names.")
+class QuotedNameArgumentTest(_QuotedNameArgumentTest):
+    pass
```
IlyaFaer, line 572: This test tries to create tables, which name contain quotes. It's not supported and gives a syntax error even if the name was escaped.

https://cloud.google.com/spanner/docs/data-definition-language#naming_conventions

AVaksman (approved):

PR #56, opened 2021-04-20
fix: composite keys reflection

IlyaFaer:


test/test_suite.py
```diff
@@ -765,6 +771,27 @@ def test_numeric_reflection(self):
...
+        reflected correctly, without considering their order.
+        """
+        # test for issue #5661
+        meta = self.metadata
+        insp = inspect(meta.bind)
+        foreign_keys = insp.get_foreign_keys(self.tables.tb2.name)
+        eq_(len(foreign_keys), 1)
+        fkey1 = foreign_keys[0]
+        eq_(set(fkey1.get("referred_columns")), {"name", "id", "attr"})
+        eq_(set(fkey1.get("constrained_columns")), {"pname", "pid", "pattr"})
```
IlyaFaer, line 792: The only diverge from the original test is that here I'm using `set` instead of a `list` - to ignore the columns order

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -481,8 +481,8 @@ def get_foreign_keys(self, connection, table_name, schema=None, **kw):
     tc.constraint_name,
     ctu.table_name,
     ctu.table_schema,
-    ccu.column_name,
-    kcu.column_name
+    ARRAY_AGG(DISTINCT ccu.column_name),
+    ARRAY_AGG(kcu.column_name)
```
IlyaFaer, line 485: A key can include several columns, which wasn't considered earlier

IlyaFaer:
The PR fixes a reflection error, so that one more test is now passing. Only single one test is still failing in `reflection` test group.

Безымянный

AVaksman (approved):

PR #57, opened 2021-04-21
feat: set sequence as an unsupported feature

IlyaFaer:


IlyaFaer:
Overriding an older [PR](https://github.com/cloudspannerecosystem/python-spanner-sqlalchemy/pull/45).

First of all, skipping the test group doesn't tell user that dialect doesn't support sequences. To mention it on the dialect level a property in `Requirements` should be set as `closed`. When the feature set as unsupported this way, the tests will also be skipped, so we don't need to import them, override and explicitly skip.

Безымянный

AVaksman (approved):

PR #59, opened 2021-04-23
test: check if a test instance exists before creating it

IlyaFaer:


IlyaFaer:
@skuruppu, I've added a check for instance existance, the workflow working fine now.

Looks like a couple of tests failed - taking a closer look at them. I don't have such a failures locally, so most likely it's just a small mistakes.

I also think we now can delete GitHub workflow with emulator - it's not working fine anyway.

skuruppu (approved):
Thanks @IlyaFaer. This is fine but there might be issues if you're running CI across multiple PRs. So as a later step, I would create unique instances per test and then have a loop to delete stale instances at the start of the test (ask @AVaksman on how we do it for Node.js).

PR #60, opened 2021-04-26
fix: avoid escaping collation name

IlyaFaer:


google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -103,6 +104,19 @@ def __init__(self, dialect):
...
+        if lc_value == '"unicode"':  # don't escape default Spanner colation
+            return False
+
+        return (
+            lc_value in self.reserved_words
+            or value[0] in self.illegal_initial_characters
+            or not self.legal_characters.match(util.text_type(value))
+            or (lc_value != value)
+        )
+
```
IlyaFaer, line 119: When I've added an identifiers preparer, it started to escape all the identifiers. The problem is that Spanner requires a collation name to be double-quoted. And if we're using a double-quoted string, it is quoted one more time by id preparer. I've tried several ways, and looks like overriding the preparer method and adding `"unicode"` as an exclusion, is the only way to fix the problem.

Безымянный

larkee (approved):
LGTM :+1:

PR #61, opened 2021-04-27
test: skip composite grouping tests

IlyaFaer:


test/test_suite.py
```diff
@@ -823,6 +824,11 @@ class IsOrIsNotDistinctFromTest(_IsOrIsNotDistinctFromTest):
     pass
 
 
+@pytest.mark.skip("Spanner doesn't support composed GROUP BY")
+class OrderByLabelTest(_OrderByLabelTest):
+    pass
```
IlyaFaer, line 829: The test is trying to do something like:
```sql
select col1 + col2
...
group by col1 + col2
```
Spanner however requires `col1` and `col2` to be grouped by or aggregated.

larkee (approved):

PR #62, opened 2021-04-27
test: fix insert test

HemangChothani:


AVaksman (approved):

test/test_suite.py
```diff
@@ -817,6 +817,10 @@ def test_insert_from_select_autoinc_no_rows(self):
     def test_insert_from_select_with_defaults(self):
         pass
 
+    @pytest.mark.skip("Spanner doesn't support auto increment")
+    def test_autoclose_on_insert(self):
+        pass
```
IlyaFaer, line 822: This test is flaky, it sometime passes. That means it can be fixed instead of skipping.

Безымянный

HemangChothani: I have tried multiple times, but it failed every time, the test is related to auto increment which spanner doesn't support. To pass this need to fill id manually which break the meaning of test.

HemangChothani: I don't think its flaky as in create table sql define `NOT NULL` to id column. `CREATE TABLE autoinc_pk (id INT64 NOT NULL, data STRING(50)\n) PRIMARY KEY (id)`. Hence override the test and also enable and override the other related advance test.

larkee (approved):

PR #65, opened 2021-05-04
ci: delete old instances

AVaksman:
This PR introduces

- Create instance with unique instance.
- Add labels including time of instance creation.
- Clean old instances erroneously left from failed system test run

### DIFF
--- a/create_test_database.py
+++ b/create_test_database.py
@@ -15,22 +15,62 @@
 # limitations under the License.
 
 import os
+import time
 
 from google.cloud.spanner_v1 import Client
+from google.cloud.spanner_v1.instance import Instance
+
+
+use_emulator = os.getenv("SPANNER_EMULATOR_HOST") is not None
 
 project = os.getenv(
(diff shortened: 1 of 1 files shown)


create_test_database.py
```diff
@@ -15,22 +15,62 @@
...
 from google.cloud.spanner_v1 import Client
+from google.cloud.spanner_v1.instance import Instance
+
+
+use_emulator = os.getenv("SPANNER_EMULATOR_HOST") is not None
 
 project = os.getenv(
     "GOOGLE_CLOUD_PROJECT", os.getenv("PROJECT_ID", "emulator-test-project"),
 )
+client = None
```
larkee, line 29: nit: `use_emulator`, `project`, and `client` are effectively global constants so they should be fully capitalized i.e. `USE_EMULATOR`, `PROJECT`, and `CLIENT`

AVaksman: Done
Thank you

larkee (approved):

PR #66, opened 2021-05-04
test: skip unsupported features testing

IlyaFaer:


IlyaFaer:
All the tests of the `select` group are processed, so adding the group into the testing workflow.

larkee:
`test_limit_offset_selectable_in_unions` is failing with the following exception:
```
google.api_core.exceptions.InvalidArgument: 400 Syntax error: Expected end of input but got keyword WHERE [at 8:12]\n LIMIT @a4 WHERE 1=1\n
```
A similar exception came up in #47 and it seems to be a bug in `spanner_dbapi` which dummy `WHERE` clauses are being mistakenly added. I've filed an issue in the `python-spanner` repo for this:  https://github.com/googleapis/python-spanner/issues/347

I am happy to keep the skips in for now with a reference to the aforementioned bug so we know to remove them once it is fixed :+1:

larkee (commented):
Both `test_limit_offset_selectable_in_unions` and `test_order_by_selectable_in_unions` are failing with the following exceptions:
```
google.api_core.exceptions.InvalidArgument: 400 Syntax error: Expected end of input but got keyword WHERE [at 8:12]\n LIMIT @a4 WHERE 1=1\n
```
```
google.api_core.exceptions.InvalidArgument: 400 Syntax error: Expected end of input but got keyword WHERE [at 6:12]\n LIMIT @a2 WHERE 1=1\n
``

A similar exception came up in #47 and it seems to be a bug in `spanner_dbapi` which dummy `WHERE` clauses are being mistakenly added. I've filed an issue in the `python-spanner` repo for this:  https://github.com/googleapis/python-spanner/issues/347

I am happy to keep the skips in for now with a reference to the aforementioned bug so we know to remove them once it is fixed :+1:

File: test/test_suite.py [hunk 1]
```diff
@@ -937,6 +937,16 @@ def test_autoclose_on_insert(self):
         assert not r.returns_rows
 
 
+class CompoundSelectTest(_CompoundSelectTest):
+    @pytest.mark.skip("Spanner doesn't support ORDER BY clause in UNION subqueries")
+    def test_limit_offset_selectable_in_unions(self):
+        pass
+
+    @pytest.mark.skip("Spanner doesn't support ORDER BY clause in UNION subqueries")
```
larkee, line 945: Please update these skip reasons to point to the previously mentioned bug if you still want to merge this PR.

I can confirm that both tests pass locally when the regex is fixed.

IlyaFaer: Oh, you already fixes the regex?!
No, no, if there is a way to fix the error, I'll just close this PR.

PR #36, opened 2021-03-25
test: add text compliance tests

HemangChothani:

PR #67, opened 2021-05-05
test: tweak test data for the Text type

IlyaFaer:


test/test_suite.py
```diff
@@ -843,3 +844,22 @@ def test_nolength_binary(self):
...
+        which is used by the original test. Overriding the test data
+        creation method to disable autoincrement and make id column
+        nullable.
+        """
+        Table(
+            "text_table",
+            metadata,
+            Column("id", Integer, primary_key=True, nullable=True),
+            Column("text_data", Text),
+        )
```
IlyaFaer, line 865: The test is actually testing `Text` type, so I suppose we can make the id column (which is `Integer` and not directly considered by the test) nullable and without autoincrement (which is not supported anyway). The change makes three  more tests passing (there are some cleanup/teardown errors, but the tests themselves are working fine - we'll take a look at cleanups little but later).

Безымянный

HemangChothani: I think https://github.com/cloudspannerecosystem/python-spanner-sqlalchemy/pull/36 PR will cover all text type tests.

IlyaFaer: #36 does too many overrides. You only need to change several symbols to add manual primary key, but it requires to copy the whole test source code for every failing test.

Plus to this, #36 includes some unclear changes, like this one:

Безымянный

It looks pretty much like a meaningful diverge from the original test. Instead all of it just a single test data override can be done, keeping the original tests as they were.

larkee (commented):
Please revert the changes to test class ordering. Additionally, could this logic be extended to other test classes or is it unique to `TextTest`?

IlyaFaer:
> Please revert the changes to test class ordering. Additionally, could this logic be extended to other test classes or is it unique to `TextTest`?

Not sure what does it mean "to test class ordering".

Yeah, I suppose there is a couple of more test cases, which could use the similar data definition override instead of overriding every test - that's what I was talking about earlier. Still, all of the PRs now merged, so it's not very obvious on which tests this logic can be extended. Requires a local revert and closer look.

larkee:
> Not sure what does it mean "to test class ordering".

Ah, my wording was unclear. The test classes have been reordered e.g. `QuotedNameArgumentTest` and `StringTest` have both been moved to a different place. Please revert this reordering so the diff shows the changes more clearly unless there was a reason for the reordering.
 
> Yeah, I suppose there is a couple of more test cases, which could use the similar data definition override instead of overriding every test - that's what I was talking about earlier. Still, all of the PRs now merged, so it's not very obvious on which tests this logic can be extended. Requires a local revert and closer look.

This looks to have been address in #75 so once the reordering has been addresses this LGTM :+1:

IlyaFaer:
@larkee, oh, I got it. Reverted.

larkee (approved):

PR #69, opened 2021-05-14
test: add a migration nox testing session

IlyaFaer:
Adding a test session with a minimal migration.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -18,11 +18,52 @@ from __future__ import absolute_import
 
 import nox
 
+ALEMBIC_CONF = """
+[alembic]
+script_location = test_migration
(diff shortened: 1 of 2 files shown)


IlyaFaer:
If something fails during the migration the whole nox session will fail as well.

IlyaFaer:
Oh, that's something new
```File "/home/circleci/project/.nox/migration_test/lib/python3.8/site-packages/alembic/ddl/impl.py", line 80, in get_by_dialect
    return _impls[dialect.name]
KeyError: 'spanner'```
I didn't yet see such an error. Working on it...

larkee (approved):
I'm not super familiar with Alembic but it LGTM :+1:

PR #70, opened 2021-05-18
test: drop alembic tables after migration test is over

IlyaFaer:


noxfile.py
```diff
@@ -157,3 +158,12 @@ def migration_test(session):
...
     shutil.rmtree("test_migration")
+
+    client = Client()
+
+    instance = client.instance("sqlalchemy-dialect-test")
+    database = instance.database("compliance-test")
+
+    database.update_ddl(["DROP TABLE account", "DROP TABLE alembic_version"]).result(
+        120
+    )
```
IlyaFaer, line 169: It's not very easy to make alembic delete its own service table, so doing it with a common Spanner objects

larkee (approved):
LGTM as long as the CI passes

PR #71, opened 2021-05-20
test: add back emulator test workflow

skuruppu:


skuruppu:
@IlyaFaer would you be able to take this PR and skip the relevant tests to get the workflow green?

I tried to see if I can debug it but I'm not sure where the primary key test is failing :(

IlyaFaer:
@skuruppu, yeah, I'm taking a look.

IlyaFaer:
@googlebot I consent.

skuruppu:
I ran some tests (using the Node.js client) against the emulator and I was able to successfully create a table with no primary key, insert a row and query it back. So I don't think that's it.

test/test_suite.py
```diff
@@ -1008,6 +1009,10 @@ def test_text_roundtrip(self):
         eq_(row, ("some text",))
 
 
+@pytest.mark.skipif(
+    os.getenv("SPANNER_EMULATOR_HOST"),
+    reason="Numeric type isn't supported by emulator yet",
+)
```
skuruppu, line 1015: I don't think these should fail. The latest release of the emulator does support NUMERIC. See https://github.com/googleapis/nodejs-spanner/pull/1358 where we enabled the tests on Node.js.

IlyaFaer:
> I ran some tests (using the Node.js client) against the emulator and I was able to successfully create a table with no primary key, insert a row and query it back. So I don't think that's it.

Unfortunately, I'm about to disagree. The first test case failing bad is `ComponentReflectionTest`. I've taken a look at its setups, and it appeared it uses two tables without primary keys. I've tried to override the setup method to explicitly set primary keys for those tables, and now `ComponentReflectionTest` passing fine.

An SQLs without primary key explicitly defined look like this in SQLAlchemy:
```sql
CREATE TABLE Singers (
            SingerId     INT64 NOT NULL,
            FirstName    STRING(1024),
            LastName     STRING(1024),
            SingerInfo   BYTES(MAX)
        ) PRIMARY KEY ()
```
Thus, primary key definition is empty. And seems like it's confusing the emulator.

AVaksman:
@googlebot I consent.

HemangChothani:
@googlebot I consent.

larkee (approved):
LGTM

IlyaFaer:
Several tests are failing with `400 Unable to infer type for parameter a1`. The error is going to be fixed in a separate PR for DB API: https://github.com/googleapis/python-spanner/pull/200
Everything else seems to be working fine. Merging.

PR #74, opened 2021-05-21
test: add performance benchmark comparator

IlyaFaer:


IlyaFaer:
Average timing got from this test suite is this:
```python
'insert_many_rows': {
    'SQLAlchemy deviation': -3.13,
    'SQLAlchemy to Spanner, %': 91,
    'SQLAlchemy, sec': 33.4,
    'Spanner, sec': 36.53
}
'insert_one_row_with_fetch_after': {
    'SQLAlchemy deviation': -0.6,
    'SQLAlchemy to Spanner, %': 66,
    'SQLAlchemy, sec': 1.18,
    'Spanner, sec': 1.78
}
'read_one_row': {
    'SQLAlchemy deviation': -0.03,
    'SQLAlchemy to Spanner, %': 92,
    'SQLAlchemy, sec': 0.33,
    'Spanner, sec': 0.36
}
'select_many_rows': {
    'SQLAlchemy deviation': -0.01,
    'SQLAlchemy to Spanner, %': 97,
    'SQLAlchemy, sec': 0.33,
    'Spanner, sec': 0.34
}
```

test/benchmark.py
```diff
@@ -0,0 +1,263 @@
...
+        conn.close()
+
+    def run(self):
+        """Execute every test case."""
+        measures = {}
+        for method in (
+            self.insert_one_row_with_fetch_after,
+            self.read_one_row,
+            self.insert_many_rows,
+            self.select_many_rows,
```
skuruppu, line 100: I'm curious to know if it's possible to choose between different transaction types and if so, then should we add tests for that. E.g. EF had `SelectMultipleSingersInReadOnlyTransactionEF` and `SelectMultipleSingersInReadWriteTransactionEF`.

skuruppu:
@IlyaFaer and @larkee any ideas on why SQLAlchemy is faster on all of these operations than Spanner?

Also, in terms of computing the average, how many times did you run it? It'll be good to have the results in the form that was presented for EF (https://github.com/cloudspannerecosystem/dotnet-spanner-entity-framework/pull/63).

test/benchmark.py
skuruppu: Also, do any of these tests trigger the checksum computation in dbapi? If not, is it possible to craft a test that does so? E.g. in autocommit mode.

test/benchmark.py
```diff
@@ -0,0 +1,263 @@
...
+
+    Insert 100 rows into a database.
+    """
+    total_count = 0
+    for row in many_rows:
+        count = transaction.execute_update(
+            "INSERT Singers (id, first_name, last_name, birth_date, picture) "
+            " VALUES {}".format(str(row))
+        )
+        total_count += count
```
skuruppu, line 238: I think if this was changed to using the [batch mutations API](https://cloud.google.com/spanner/docs/getting-started/python#write-data-with-mutations), then there might be a noticeable difference between the spanner client and SQLAlchemy. Can you also run a test for that to see what the difference is? Because we will have to support it for GA even if we don't for preview.

test/benchmark.py
skuruppu: You can ignore my first comment. I found the discussion about this in the chat.

IlyaFaer:
> @IlyaFaer and @larkee any ideas on why SQLAlchemy is faster on all of these operations than Spanner?
> 
> Also, in terms of computing the average, how many times did you run it? It'll be good to have the results in the form that was presented for EF ([cloudspannerecosystem/dotnet-spanner-entity-framework#63](https://github.com/cloudspannerecosystem/dotnet-spanner-entity-framework/pull/63)).

I've executed the tests in circle of 15 repetitions for the last time, and calculated the average time. Would do more reps, but it takes time, and I've been working on another task in parallel. Yeah, I'll take a look at EFs way of calculating...

Well, I'd not say that SQLAlchemy is faster, the difference is not even 10%. Most likely it relates to some internal things, like SQLAlchemy does `execute_many()`, while Spanner does `execute()` with a single request, where all the args are compiled into single args list and so on.

test/benchmark.py
```diff
@@ -0,0 +1,263 @@
...
+
+    if total_count != 99:
```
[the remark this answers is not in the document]
IlyaFaer: The thing is that we don't track number of inserted rows in DB API. Not sure why, maybe we should?

test/benchmark.py
IlyaFaer: > Also, do any of these tests trigger the checksum computation in dbapi? If not, is it possible to craft a test that does so? E.g. in autocommit mode.

Checksums are calculated in !autcommit mode, which is the default mode, so tests should already be calculating checksums.

test/benchmark.py
IlyaFaer: Yeah, that's true, here are first measures:
```python
{'insert_many_rows_with_mutations': {
    'SQLAlchemy deviation': 33.33,
    'SQLAlchemy to Spanner, %': 2644,
    'SQLAlchemy, sec': 34.64,
    'Spanner, sec': 1.31
}}
```
Hm-m, I'll take a look if we can override something somewhere to make SQLAlchemy execute inserts with mutations instead of `Cursor.execute_many()`, but no promises yet.

IlyaFaer:
Here are the results from 50 runs with statistical calculations:

|Test|mean|error|std_dev|
|----|-------|-----|--------|
|SPANNER insert_one_row_with_fetch_after| 1.79|0.01|0.1|
|ALCHEMY insert_one_row_with_fetch_after|  1.07| 0.0|0.02|
|SPANNER read_one_row| 0.33| 0.0| 0.01|
|ALCHEMY read_one_row| 0.29| 0.0| 0.01|
|SPANNER insert_many_rows|  16.23| 0.06| 0.39|
|ALCHEMY insert_many_rows|  32.33| 0.08| 0.53|
|SPANNER select_many_rows| 0.36| 0.01| 0.08|
|ALCHEMY select_many_rows|   0.19| 0.0| 0.01|
|SPANNER insert_many_rows_with_mutations| 0.33| 0.0| 0.03|
|SQLALCHEMY insert_many_rows_with_mutations| 32.1| 0.07| 0.49|

Strange, but I have very stormy raw data for `SPANNER insert_many_rows`. Previously it was giving me ~30 sec, but now it's mostly ~15 sec. The only thing that I've done is a pull in python-spanner repo. Something was optimized recently?!

SQLAlchemy `insert_many_rows` is significantly slower than Spanner's one, as SQLAlchemy do `execute_many()`. Will take a look if we can optimize something there.

test/benchmark.py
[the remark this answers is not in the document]
larkee: I believe you can get the number of inserted rows for an executed statement via [`cursor.rowcount`](https://github.com/googleapis/python-spanner/blob/master/google/cloud/spanner_dbapi/cursor.py#L115-L122).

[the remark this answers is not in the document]
IlyaFaer: Right now `rowcount` only reflects number of read rows and updated. Insert command results are not processed in any way.

larkee:
Getting the following error when I run the benchmark locally:
```
Traceback (most recent call last):
  File "test/benchmark.py", line 308, in <module>
    spanner_stats[key + "_aver"] = round(statistics.mean(value), 2)
  File "/usr/local/google/home/pottinger/.pyenv/versions/3.8-dev/lib/python3.8/statistics.py", line 315, in mean
    raise StatisticsError('mean requires at least one data point')
statistics.StatisticsError: mean requires at least one data point
```

I'm trying to figure out why SQLAlchemy is faster than the Spanner client

IlyaFaer:
> Getting the following error when I run the benchmark locally:
> 
> ```
> Traceback (most recent call last):
>   File "test/benchmark.py", line 308, in <module>
>     spanner_stats[key + "_aver"] = round(statistics.mean(value), 2)
>   File "/usr/local/google/home/pottinger/.pyenv/versions/3.8-dev/lib/python3.8/statistics.py", line 315, in mean
>     raise StatisticsError('mean requires at least one data point')
> statistics.StatisticsError: mean requires at least one data point
> ```

If an error occured while running a test, its time is not considered and is erased from the list of measures. So, I guess something's failing in your case, most likely `Aborted` in case of inserting many rows.

> I'm trying to figure out why SQLAlchemy is faster than the Spanner client

Yeah, and I'm trying to speed up `executemany()` for insert calls.

IlyaFaer:
@larkee, pushing the session initiation code.

I've executed tests only in 7-times circle yet, but it's seen that Spanner cases became faster.

test/benchmark.py
```diff
@@ -0,0 +1,327 @@
...
+        raise ValueError("Received invalid last name: " + last_name)
+
+
+def insert_many_rows(transaction, many_rows):
+    """A transaction-function for the original Spanner client.
+
+    Insert 100 rows into a database.
+    """
+    total_count = 0
+    for row in many_rows:
```
larkee, line 268: This can be simplified to use `BatchDML` like so:
```
statements = [
    "INSERT Singers (id, first_name, last_name, birth_date, picture) VALUES {}".format(str(row)) for row in many_rows
]
counts = transaction.execute_update(statements)
total_count = sum(counts)
```

test/benchmark.py
```diff
@@ -0,0 +1,327 @@
...
+        """Execute the wrapped method and measure its execution time.
+
+        Args:
+            measures (dict): Test cases and their execution time.
+        """
+        t_start = time.time()
+        try:
+            function(self)
+            measures[function.__name__] = round(time.time() - t_start, 2)
+        except Exception:
```
larkee, line 50: Catching all exceptions is bad practice as it makes it difficult to know when things go wrong. I would much prefer you catch on specific error such as `Aborted` or `DeadlineExceeded` or whatever. Currently, if there is an issue relating to a table or row already existing, there is no indication of this.

larkee:
For anyone looking at the above results: 

The difference in time for `insert_one_row_with_fetch_after` was caused by session creation being included in the timing for the Spanner client but not for SQLAlchemy. When the session creation is done in advance, the Spanner client is faster. 

`SPANNER insert_many_rows` is not using `BatchDML` which is much faster. This will be updated.

I believe the timing difference between the read and select comes from the Spanner client returning the active session to the session pool while SQLAlchemy keeps it. However I need to look into this further to confirm.

test/benchmark.py
IlyaFaer: I suspect you mean `batch_update()`. Okay, done

test/benchmark.py
IlyaFaer: Well, as this is a performance benchmark, not a functionality test, I've decided to catch all the errors. However, true, it can cause some troubles. I think the only annoying exception right now is `Aborted`. Will see if others will appear.

IlyaFaer:
@larkee, I'll try to rerun the statistical benchmark to see how the average time for tens of tests changed with all of the recent pushes.

IlyaFaer:
New measures:

|Test|mean|error|std_dev|
|----|-------|-----|--------|
|SPANNER insert_one_row_with_fetch_after| 0.91|0.01|0.09|
|ALCHEMY insert_one_row_with_fetch_after|  1.07| 0.0|0.03|
|SPANNER read_one_row| 0.33| 0.0| 0.01|
|ALCHEMY read_one_row| 0.3| 0.0| 0.01|
|SPANNER insert_many_rows|  1.37| 0.02| 0.12|
|ALCHEMY insert_many_rows|  32.38| 0.09| 0.53|
|SPANNER select_many_rows| 0.31| 0.02| 0.08|
|ALCHEMY select_many_rows|   0.22| 0.03| 0.01|
|SPANNER insert_many_rows_with_mutations| 0.34| 0.0| 0.03|
|SQLALCHEMY insert_many_rows_with_mutations| 32.2| 0.09| 0.49|

test/benchmark.py
```diff
@@ -0,0 +1,327 @@
...
+
+
+class SpannerBenchmarkTest(BenchmarkTestBase):
+    """The original Spanner performace testing class."""
+
+    def __init__(self):
+        super().__init__()
+        self._client = Client()
+        self._instance = self._client.instance("sqlalchemy-dialect-test")
+        self._database = self._instance.database("compliance-test2")
```
larkee, line 122: The base class is creating the `compliance-test` database. From what I can tell `compliance-test2` isn't referenced anywhere else.
```suggestion
        self._database = self._instance.database("compliance-test")
```

test/benchmark.py
```diff
@@ -0,0 +1,327 @@
...
+            batch.insert(
+                table="Singers",
+                columns=("id", "first_name", "last_name", "birth_date", "picture"),
+                values=self._many_rows2,
+            )
+
+    @measure_execution_time
+    def read_one_row(self):
+        with self._database.snapshot() as snapshot:
+            keyset = KeySet(keys=([1],))
```
larkee, line 157: In the interest of being equivalent, this should probably be:
```suggestion
            keyset = KeySet(all_=True)
```
The SQLAlchemy SQL is executing `SELECT * FROM Singers` while I think this is currently executing the equivalent of `SELECT * FROM Singers WHERE id = 1`

larkee (approved):
LGTM other than some minor nits

test/benchmark.py
IlyaFaer: Oh. Locally I run tests on `compliance_test2` - it comes from times when we were using only one instance. This line slipped through my radars.

PR #75, opened 2021-05-24
refactor: remove unnecessary override methods

HemangChothani:


larkee (commented):
LGTM, although aren't there overrides in `NumericTest` and `TextTest` which would be similarly refactored?

HemangChothani:
For `TextTest` covered in PR https://github.com/cloudspannerecosystem/python-spanner-sqlalchemy/pull/67 and need to override `NumericTest` tests as don't have any other option to resolve the limitation of spanner that a table with no primary key columns can have only one row.

larkee (approved):

PR #81, opened 2021-05-27
test: generate dynamic db url  for test

HemangChothani:


create_test_database.py
```diff
@@ -72,5 +74,17 @@ def prep_instance():
...
 
+    config = configparser.ConfigParser()
+
+    config.read("setup.cfg")
+    url = "spanner:///projects/{project}/instances/{instance_id}/databases/compliance-test".format(
+        project=PROJECT, instance_id=instance_id
+    )
+    config.set("db", "default", url)
+
+    with open("setup.cfg", "w") as configfile:
```
HemangChothani, line 85: Another solution is to create `test.cfg` file and write db_url every time which override the db tag of `setup.cfg ` file.

AVaksman: I do like the option of creating a `test.cfg` file where [db] -> default URL will be dynamically updated with every test run.

larkee (approved):
LGTM.

nit: should use `test:` instead of `fix:`

Issue #79, opened 2021-05-26
How to force using a secondary index


hadim:
The below query using SQLAlcehmy correctly use the secondary index implicitly:

```sql
# NO TABLE FULL SCAN
SELECT
    *
FROM
    table_1
WHERE
    table_1.int_idx IN (
        421258595511884968,
        547782519786340766,
        129301010718819286,
        166700487714791022,
        342426916160168051,
        138424151540458013,
        284631122917766146,
        2004816598370226,
        8860956766006702,
        102089802699046030,
        448788799769600799,
        238663649879175555,
        160112916841312022,
        327957770675106772,
        526038790686069490,
        488214752010784845,
        248198921755348720,
        881388131744695032,
        500160396744685082,
        334271611984959358
    )
```

Now if I want to join it with an interleaved table with:

```sql
# TRIGGER TABLE FULL SCAN
SELECT
  *
FROM
  table_1
JOIN
  table_2
ON
  table_1.primary_key = table_2.primary_key
WHERE
  table_1.int_idx IN (421258595511884968,
    547782519786340766,
    129301010718819286,
    166700487714791022,
    342426916160168051,
    138424151540458013,
    284631122917766146,
    2004816598370226,
    8860956766006702,
    102089802699046030,
    448788799769600799,
    238663649879175555,
    160112916841312022,
    327957770675106772,
    526038790686069490,
    488214752010784845,
    248198921755348720,
    881388131744695032,
    500160396744685082,
    334271611984959358)
```

It triggers a table full scan. To make it efficient I can specify the secondary index manually in the query and it works just well:

```sql
# NO TABLE FULL SCAN
SELECT
  *
FROM
  table_1@{FORCE_INDEX=table_1_by_int_idx}
JOIN
  table_2
ON
  table_1.primary_key = table_2.primary_key
WHERE
  table_1.int_idx IN (421258595511884968,
    547782519786340766,
    129301010718819286,
    166700487714791022,
    342426916160168051,
    138424151540458013,
    284631122917766146,
    2004816598370226,
    8860956766006702,
    102089802699046030,
    448788799769600799,
    238663649879175555,
    160112916841312022,
    327957770675106772,
    526038790686069490,
    488214752010784845,
    248198921755348720,
    881388131744695032,
    500160396744685082,
    334271611984959358)
```

[...]

PR #82, opened 2021-05-28
feat: support table hints

IlyaFaer:
Towards #79

### DIFF
--- a/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
+++ b/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
@@ -125,6 +125,16 @@ class SpannerSQLCompiler(SQLCompiler):
 
     compound_keywords = _compound_keywords
 
+    def get_from_hint_text(self, _, text):
+        """Return a hint text.
+
+        Overriden to avoid adding square brackets to the hint text.
+
+        Args:
+            text (str): The hint text.
+        """
+        return text
+
     def visit_empty_set_expr(self, type_):
         """Return an empty set expression of the given type.
 
--- a/test/test_suite.py
+++ b/test/test_suite.py
@@ -1189,10 +1189,18 @@ class IsOrIsNotDistinctFromTest(_IsOrIsNotDistinctFromTest):
     pass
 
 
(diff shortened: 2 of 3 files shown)


google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -125,6 +125,16 @@ class SpannerSQLCompiler(SQLCompiler):
...
+    def get_from_hint_text(self, _, text):
+        """Return a hint text.
+
+        Overriden to avoid adding square brackets to the hint text.
+
+        Args:
+            text (str): The hint text.
+        """
+        return text
+
```
IlyaFaer, line 137: The original method does the same, but also adds square brackets without conditions and explanations

test/test_suite.py
```diff
@@ -1194,19 +1198,46 @@ def test_group_by_composed(self):
...
-        "Spanner DBAPI incorrectly classify the statement starting with brackets."
-    )
-    def test_order_by_selectable_in_unions(self):
-        pass
+    def test_complex_query_table_hints(self):
+        EXPECTED_QUERY = (
+            "SELECT users.id, users.name \nFROM users @{FORCE_INDEX=table_1_by_int_idx}"
+            " JOIN addresses ON users.id = addresses.user_id "
+            "\nWHERE users.name IN (%s, %s)"
+        )
```
IlyaFaer, line 1214: The original query with a correct hint

hadim:
Just tested it and it works fine for me with a query very similar to the one you have this PR. Thanks!

test/test_suite.py
```diff
@@ -1194,19 +1198,46 @@ def test_group_by_composed(self):
         pass
 
 
-class CompoundSelectTest(_CompoundSelectTest):
```
larkee, line 1197: To confirm, the `CompoundSelectTest` overrides are being removed because the DBAPI fix has been applied.

I would prefer a separate PR to remove these overrides because I'm 90% sure there are other tests with this skip reason.

larkee (approved):
LGTM other than the nit regarding `CompoundSelectTest`

PR #84, opened 2021-06-01
test: add join tests

HemangChothani:


test/test_suite.py
```diff
@@ -1210,3 +1210,110 @@ def test_limit_offset_selectable_in_unions(self):
     )
     def test_order_by_selectable_in_unions(self):
         pass
+
+
+class JoinTest(fixtures.TablesTest):
```
larkee, line 1215: Why are we reimplementing this test class rather than overloading it?

larkee: Aren't they part of `test_select`: https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/testing/suite/test_select.py#L625-L722

PR #85, opened 2021-06-02
feat: support interleaved tables feature

IlyaFaer:
Adding two dialect-specific arguments:
`spanner_interleave_in` - a name of a parent table
`spanner_inverleave_on_delete_cascade` - a flag indicating if `ON DELETE CASCADE` statement should be added for this inverleave relation

### DIFF
--- a/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
+++ b/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
@@ -270,8 +270,13 @@ class SpannerDDLCompiler(DDLCompiler):
             str: primary key difinition to add to the table CREATE request.
         """
         cols = [col.name for col in table.primary_key.columns]
+        post_cmds = " PRIMARY KEY ({})".format(", ".join(cols))
 
-        return " PRIMARY KEY ({})".format(", ".join(cols))
+        if table.kwargs.get("spanner_interleave_in"):
+            post_cmds += ",\nINTERLEAVE IN PARENT {} ON DELETE CASCADE".format(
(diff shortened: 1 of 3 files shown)


IlyaFaer:
@larkee, all the commits showed up, and the PR is ready for review.

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -287,15 +287,27 @@ def visit_unique_constraint(self, constraint):
     def post_create_table(self, table):
         """Build statements to be executed after CREATE TABLE.
 
+        Includes "primary key" and "interleaved
+        table" statements generation.
```
skuruppu, line 291:
```suggestion
        Includes "primary key" and "interleaved table" statements generation.
```

test/test_suite.py
```diff
@@ -1436,3 +1437,58 @@ class Address(Base):
         query = query.join(Address)
 
         assert str(query.statement.compile(session.bind)) == EXPECTED_QUERY
+
+
+class InterleavedTablesTest(fixtures.TestBase):
+    """
+    Check that CREATE TABLE statements for
+    interleaved tables are correctly generated.
```
skuruppu, line 1445:
```suggestion
    Check that CREATE TABLE statements for interleaved tables are correctly
    generated.
```

skuruppu (approved):
This looks good to me. I would also like @olavloite to take a look in case there are cases that need to be handled that we haven't thought of.

olavloite (approved):

PR #88, opened 2021-06-11
fix: for open-telemetry status code spec change

### LINKED ISSUE #89: Update open telemetry tracing code according to new spec.
The method used in [open telemetry tracing](https://github.com/cloudspannerecosystem/python-spanner-sqlalchemy/blob/main/google/cloud/sqlalchemy_spanner/_opentelemetry_tracing.py#L28) has been updated in this [commit]( https://github.com/lzchen/opentelemetry-python/commit/cf13537eaa8f6159598b5c17e1159b9571a3996c#diff-af63e517bdd2a33c2a8833eb083e0b0f640e0c62a591f0d289deb42d5cb3ce60R25)

## Steps to Reproduce the Problem

```python
from opentelemetry.trace.status import Status, StatusCanonicalCode    
```

## Stack trace
```
ImportError: cannot import name 'StatusCanonicalCode' from 'opentelemetry.trace.status'
```

Stackoverflow [question link with more context](https://stackoverflow.com/questions/67775140/getting-import-error-while-trying-to-import-statuscanonicalcode-using-openteleme)

## Specifications
opentelemetry-api                    1.2.0
opentelemetry-sdk                   1.2.0
google-cloud-spanner             3.4.0

vi3k6i5:
Updated support for open telemetry tracing and updated setup.py extra dependency.
Also added unit tests for open telemetry.

fixes #89


test/_helpers.py
```diff
@@ -0,0 +1,80 @@
...
+        return
+
+    provider = trace.get_tracer_provider()
+    if not hasattr(provider, "add_span_processor"):
+        return
+    provider.add_span_processor(SimpleSpanProcessor(get_test_ot_exporter()))
+    _TEST_OT_PROVIDER_INITIALIZED = True
+
+
+class OpenTelemetryBase(fixtures.TestBase):  # , unittest.TestCase):
```
larkee, line 53: Use `unittest.TestCase` as the base class. This test class is only being used for the unit tests so it does not need the SQLAlchemy fixtures.

larkee (commented):
LGTM. However, currently the new tests aren't being run. Please add a `unit` nox session to `noxfile.py` similar to [here](https://github.com/googleapis/python-spanner/blob/master/noxfile.py#L115-L130).

vi3k6i5:
> LGTM. However, currently the new tests aren't being run. Please add a `unit` nox session to `noxfile.py` similar to [here](https://github.com/googleapis/python-spanner/blob/master/noxfile.py#L115-L130).

Added the same.

noxfile.py
```diff
@@ -110,10 +110,17 @@ def lint_setup_py(session):
...
+    session.install("mock")
     session.install("-e", ".")
     session.run("python", "create_test_database.py")
     session.run("pytest", "-v")
 
+    # Run SQLAlchemy dialect compliance test suite with OpenTelemetry.
+    session.install("opentelemetry-api==1.1.0")
+    session.install("opentelemetry-sdk==1.1.0")
+    session.install("opentelemetry-instrumentation==0.20b0")
+    session.run("pytest", "-v")
```
larkee, line 122: The unit tests are using `unittest`, not `pytest`. This is just running the compliance tests again. You need to use `session.run("py.test", "--quiet", os.path.join("tests"), *session.posargs)` (or something similar). We can consider whether coverage is useful another time.

Additionally, I'd prefer the unit tests to have their own session so they can be run separately. That way if a developer is writing unit tests they don't have to wait for the compliance tests to finish running to get results.

This should be as simple as writing another function, e.g.
```
nox.session(python=DEFAULT_PYTHON_VERSION)
def unit(session):
    """Run unit tests."""
    // TODO: write code for running the unit tests
```

vi3k6i5: Separate into a new `unit` session for unit test.

noxfile.py
```diff
@@ -110,11 +111,25 @@ def lint_setup_py(session):
...
+def unit(session):
+    """Run unit tests."""
+    # Run SQLAlchemy dialect compliance test suite with OpenTelemetry.
+    session.install("pytest")
+    session.install("mock")
+    session.install("-e", ".")
+    session.install("opentelemetry-api==1.1.0")
+    session.install("opentelemetry-sdk==1.1.0")
+    session.install("opentelemetry-instrumentation==0.20b0")
+    session.run("py.test", "--quiet", os.path.join("test"), *session.posargs)
```
larkee, line 130: Strangely, this seemed to run the compliance suite tests as well 😕 Perhaps create a `test/unit` folder, move the OT tests, and update the path. That way if we add more unit test files in future it's very clear where to put them and they'll run correctly until this session 👍

vi3k6i5: Moved it to `test/unit/` Also had to change to `fixtures.TestBase` as base class for `OpenTelemetryBase` instead of `unittest.Testcase` as the later caused the tests to be skipped.

larkee (approved):

PR #94, opened 2021-06-23
test: skip null/numeric related tests on emulator

IlyaFaer:


test/test_suite.py
```diff
@@ -328,7 +329,17 @@ class DateTest(DateFixtureTest, _DateTest):
     and maintain DRY concept just inherit the class to run tests successfully.
     """
 
-    pass
+    @pytest.mark.skipif(
+        bool(os.environ.get("SPANNER_EMULATOR_HOST")), reason="Skipped on emulator"
```
larkee, line 333: nit: You don't need to explicitly cast this to a bool; Python will do it automatically.

test/test_suite.py
```diff
@@ -1110,6 +1143,9 @@ def test_numeric_as_float(self):
             [15.7563],
         )
 
+    @pytest.mark.skipif(
+        bool(os.environ.get("SPANNER_EMULATOR_HOST")), reason="Skipped on emulator"
+    )
```
larkee, line 1148: These two float tests don't contain null values so they aren't affected by the lack of null value type inference like the other tests. Additionally, they seem to be passing on the other PR:  https://github.com/cloudspannerecosystem/python-spanner-sqlalchemy/pull/85/checks?check_run_id=2849752522

So I don't think they should be skipped

test/test_suite.py
IlyaFaer: First I've tried to write the code without `bool` and it failed with `SyntaxError`. It was three commits ago:
https://github.com/cloudspannerecosystem/python-spanner-sqlalchemy/pull/94/commits/9da642c37599847a88e908e4bc9267230b9dd1a0

skuruppu (approved):

PR #102, opened 2021-08-11
Create BENCHMARKS.md

IlyaFaer:


BENCHMARKS.md
```diff
@@ -0,0 +1,14 @@
+# 07-11-2021
+
```
skuruppu, line 2: If you could also add some details on:
1. What we're comparing?
2. How to run the benchmarks (with a pointer to the file that defines the benchmarks)?
3. The specific setup for this set of results.

That way, anyone looking at this has the context of what the results mean.

skuruppu (approved):

PR #114, opened 2021-09-02
chore: create sync-repo-settings

busunkim96:
This repo does not use Kokoro for presubmits, so it needs a set of requiredStatusCheckContexts different from [the default](https://github.com/googleapis/repo-automation-bots/blob/21bd620e5e09f3bc6e98995bcd1bce6c9c9ee51a/packages/sync-repo-settings/src/required-checks.json#L45-L48) for Python.

### DIFF
--- a/.github/sync-repo-settings.yaml
+++ b/.github/sync-repo-settings.yaml
@@ -9,3 +9,4 @@ branchProtectionRules:
   requiredStatusCheckContexts:
     - 'tests'
     - 'cla/google'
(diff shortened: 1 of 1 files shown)


busunkim96:
@skuruppu Could you let me know which of the checks should be marked required?

skuruppu (approved):

skuruppu:
I think we should also mark "ci/circleci: build-and-test" as required.

Issue #115, opened 2021-09-07
feat: set user-agent string to distinguish SQLAlchemy requests


skuruppu:
**Is your feature request related to a problem? Please describe.**
The implementation doesn't seem to set a user-agent string to indicate that the requests are coming from SQLAlchemy.

**Describe the solution you'd like**
Set the user-agent in the dbapi connection.

**Describe alternatives you've considered**
N/A

**Additional context**
It seems that [dbapi](https://github.com/googleapis/python-spanner/blob/48ac924672c8387e3d47da4eace00c3bdbf3baf6/google/cloud/spanner_dbapi/version.py#L19) by default sets the `DEFAULT_USER_AGENT = "django_spanner/" + VERSION`. This may not be ideal.

IlyaFaer:
Pushed changes for SQLAlchemy: https://github.com/googleapis/python-spanner-sqlalchemy/pull/116

As I see, Django already uses its own user agent (though it's hardcoded): https://github.com/googleapis/python-spanner-django/blob/4643876219e8a54feb94bf14a79f0fe2fbe3971a/django_spanner/base.py#L135

So, the only place left is the DB API itself:
https://github.com/googleapis/python-spanner/blob/48ac924672c8387e3d47da4eace00c3bdbf3baf6/google/cloud/spanner_dbapi/version.py#L19
What name it should have? It's DB API, but it lives in `google-cloud-spanner`. Is it desired to divide the original client and the DB API user agents?

skuruppu:
Good question, I think it is good to distinguish between the client library and dbapi. So how about we default dbapi to something like "dbapi/<version>" where version is the Spanner client version as they both get released in the same package?

geudrik:
Looks like I'm just late to the party, based on @IlyaFaer 's MR that's been merged

skuruppu:
@IlyaFaer I assume you'll have a PR coming to fix the default in dbapi?

@geudrik I'm a bit torn on whether to default dbapi to have `python-spanner/<version>` or `dbapi/<version`. I think ideally we'd distinguish the Python client from dbapi since they have different use cases. But I understand that it's in the same package so maybe it's confusing to distinguish it.

geudrik:
I'm just an internet stranger, so take my input with some salt. It's your repo after all!

[...]

PR #116, opened 2021-09-07
fix: set correct user agent for the dialect

IlyaFaer:
Towards #115


google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -433,9 +434,10 @@ def create_connect_args(self, url):
             ),
             url.database,
         )
+        dist = pkg_resources.get_distribution("sqlalchemy-spanner")
         return (
             [match.group("instance"), match.group("database"), match.group("project")],
-            {},
+            {"user_agent": dist.project_name + "/" + dist.version},
```
skuruppu, line 440: If you could also add a quick unit test for this, that would be great. Just to see if it's getting set correctly.

test/test_suite.py
```diff
@@ -1525,3 +1526,23 @@ def test_interleave_on_delete_cascade(self):
...
+        self._metadata = MetaData(bind=self._engine)
+
+    def test_user_agent(self):
+        dist = pkg_resources.get_distribution("sqlalchemy-spanner")
+
+        with self._engine.connect() as connection:
+            assert (
+                connection.connection.instance._client._client_info.user_agent
+                == dist.project_name + "/" + dist.version
+            )
```
IlyaFaer, line 1548: Initiating a connection and checking that its underlying client has a correct user agent

skuruppu (approved):

PR #123, opened 2021-09-24
feat: support SQLAlchemy 1.4

IlyaFaer:
The PR contains changes necessary for SQLAlchemy 1.4 support.

As tests for 1.3 and 1.4 versions have a lot of differences (non-functionaly, only tests), they are splitted into two files `test_suite_13.py` and `test_suite_14.py`. The test suites should be executed in parallel (to save time) and under different versions of SQLAlchemy.

The PR also includes a `setup.py` file change and migration test change related to the dialect registering. It should work fine both for SQLAlchemy 1.3 and 1.4, and doesn't require package reinstallation.


IlyaFaer:
Migration test is failing here, but locally I have it passing fine both for 1.3 and 1.4 version. I suppose it's something test environment related...
image

PR #124, opened 2021-09-27
fix: ALTER COLUMN NOT NULL directive fails because of inappropriate s…

IlyaFaer:


google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -864,3 +872,28 @@ def do_execute_no_params(self, cursor, statement, context=None):
...
+# Alembic ALTER operation override
+@compiles(ColumnType, "spanner")
+def visit_column_type(
+    element: "ColumnType", compiler: "SpannerDDLCompiler", **kw
+) -> str:
+    return "%s %s %s" % (
+        alter_table(compiler, element.table_name, element.schema),
+        alter_column(compiler, element.column_name),
+        "%s" % format_type(compiler, element.type_),
+    )
```
IlyaFaer, line 899: There are two syntax diverges:
Spanner supports `NULL/NOT NULL` statement without the `DROP` word in front of it for constraints dropping
Spanner doesn't use `TYPE` word for column type mentions

These two overrides covers both items.

noxfile.py
```diff
@@ -60,6 +60,12 @@ class = StreamHandler
         sa.Column('id', sa.Integer, primary_key=True),
         sa.Column('name', sa.String(50), nullable=False),
         sa.Column('description', sa.Unicode(200)),
+    )
+    op.alter_column(
+        'account',
+        'name',
+        existing_type=sa.String(50),
+        nullable=True,
```
IlyaFaer, line 68: Adding the issue case into the migration test (the original error could be repeated this way before these changes)

noxfile.py
```diff
@@ -123,6 +130,7 @@ def unit(session):
     # Run SQLAlchemy dialect compliance test suite with OpenTelemetry.
     session.install("pytest")
     session.install("mock")
+    session.install("alembic")
```
IlyaFaer, line 133: As we're now doing some alembic overrides, we need to install it

PR #135, opened 2021-10-19
feat: support JSON data type

IlyaFaer:
Based on another PR: https://github.com/googleapis/python-spanner/pull/627


skuruppu:
Yeah JSON is not available in the emulator yet AFAIK.

skuruppu:
@IlyaFaer will you be able to take a look at the Kokoro failures?

Also, please skip the JSON tests when being run against the emulator. We don't yet have an ETA for when the emulator will support JSON.

IlyaFaer:
@skuruppu, of course, I'm actually working on those failures, just a little bit stuck with strange errors.

IlyaFaer:
@skuruppu, looks like everything is okay now.

I guess we should release the `python-spanner` with all the JSON-related changes and then update the dependencies here, in the dialect code. To make sure the dialect uses the `python-spanner` version with all the necessary changes.

skuruppu:
Thanks a lot @IlyaFaer for the fixes. We're in a release freeze right now due to the public holidays in the US. But we should be able to release the Python client on 1st Dec and then we can update the dependency here.

@larkee this is now ready for review.

larkee (commented):
LGTM but I'd like to wait on the client release next week to confirm this works as intended against prod

skuruppu:
@IlyaFaer would you please be able to take a look at the compliance test failure?

larkee (approved):

PR #139, opened 2021-11-03
feat: support computed columns

### LINKED ISSUE #137: feat: support generated columns
**Is your feature request related to a problem? Please describe.**
I was trying to setup the following table:

```
singers = Table(
    "Singers",
    metadata,
    Column("SingerId", String(36), primary_key=True, nullable=False),
    Column("FirstName", String(200)),
    Column("LastName", String(200), nullable=False),
    Column("FullName", String(400), nullable=False, Computed("COALESCE(FirstName || ' ', '') || LastName")
)
```

and I got this error from the Spanner backend:

```
google.api_core.exceptions.InvalidArgument: 400 Error parsing Spanner DDL statement: CREATE TABLE `Singers` (\n\t`SingerId` STRING(36) NOT NULL, \n\t`FirstName` STRING(200), \n\t`LastName` STRING(200) NOT NULL, \n\t`FullName` STRING(400) GENERATED ALWAYS AS (COALESCE(FirstName || \' \', \'\') || LastName) NOT NULL\n) PRIMARY KEY (SingerId) : Syntax error on line 5, column 25: Expecting \')\' but found \'GENERATED\'
```

This happens because SQLAlchemy translates `Generated("<computation>")` to `GENERATED ALWAYS AS (<computation>) STORED`. But [Spanner syntax](https://cloud.google.com/spanner/docs/generated-column/how-to) for generated colums is `AS (<computation>) STORED`.

**Describe the solution you'd like**
Support generated columns as Spanner expects it.

IlyaFaer:
Closes #137


google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -254,6 +254,13 @@ def limit_clause(self, select, **kw):
 class SpannerDDLCompiler(DDLCompiler):
     """Spanner DDL statements compiler."""
 
+    def visit_computed_column(self, generated, **kw):
+        """Computed column operator."""
+        text = "AS (%s) STORED" % self.sql_compiler.process(
+            generated.sqltext, include_table=False, literal_binds=True
+        )
+        return text
+
```
IlyaFaer, line 263: Overriding the method to drop `GENERATED ALWAYS` part of the statement

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -499,14 +506,20 @@ def get_columns(self, connection, table_name, schema=None, **kw):
...
+                }
+
+                if col[3] is not None:
+                    col_desc["computed"] = {
+                        "persisted": True,
+                        "sqltext": col[3],
                     }
-                )
+                cols_desc.append(col_desc)
+
```
IlyaFaer, line 522: Some changes of the reflection method are required, indeed like this: if the column is not generated, `computed` field shoud not exist in the result (checked by the tests).

test/test_suite.py
```diff
@@ -1594,3 +1600,87 @@ def test_read_only(self):
...
+                    )
+                )
+
+
+class ComputedReflectionTest(_ComputedReflectionTest, ComputedReflectionFixtureTest):
+    @pytest.mark.skip("Default values are not supported.")
+    def test_computed_col_default_not_set(self):
+        pass
+
+    def test_get_column_returns_computed(self):
```
larkee, line 1676: Can you add a comment explaining why this override exists?

test/test_suite.py
```diff
@@ -1594,3 +1600,87 @@ def test_read_only(self):
...
+                        "computed_stored",
+                        Integer,
+                        Computed("normal * 42", persisted=True),
+                    )
+                )
+
+
+class ComputedReflectionTest(_ComputedReflectionTest, ComputedReflectionFixtureTest):
+    @pytest.mark.skip("Default values are not supported.")
+    def test_computed_col_default_not_set(self):
```
larkee, line 1673: Just to confirm, this test is testing that the defaults are used if no values are set?

larkee (commented):
The changes look good to me. I just have a few questions about the tests.

Currently, `test_get_column_returns_persisted_with_schema` is marked as `SKIPPED`. It looks like this is because the `schemas` requirement is closed. I just want to verify that this requirement should be closed.

test/test_suite.py
IlyaFaer: Default values for computed columns

test/test_suite.py
IlyaFaer: Done

larkee (approved):

PR #146, opened 2021-11-16
feat: support stale reads

IlyaFaer:


test/test_suite.py
```diff
@@ -1594,3 +1594,13 @@ def test_read_only(self):
...
+
+    def test_staleness(self):
+        with self._engine.connect().execution_options(
+            staleness={"max_staleness": 5}
+        ) as connection:
+            connection.execute(select(["*"], from_obj=self._table)).fetchall()
+            assert connection.connection.staleness == {"max_staleness": 5}
+
+        with self._engine.connect() as connection:
+            assert connection.connection.staleness is None
```
IlyaFaer, line 1606: First I've tried to run the test without `reset_connection()` implemented to see if the `staleness` option is really kept on a connection even in pool. Yes, it is - as the pool size is `1`, the same connection will be returned on the line 1605, so the `staleness` option will be equal to the line 1600. But with `reset_connection()` hook implemented, the option is dropped when a connection is returned back to the pool.

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -38,6 +40,13 @@
 from google.cloud import spanner_dbapi
 from google.cloud.sqlalchemy_spanner._opentelemetry_tracing import trace_call
 
+
+@listens_for(Pool, "reset")
+def reset_connection(dbapi_conn, connection_record):
+    """An event of returning a connection back to a pool."""
+    dbapi_conn.connection.staleness = None
```
IlyaFaer, line 47: According to the docs, connection the hook to the `Pool` class will connect it to all the pools used by the dialect. In the test below a `QueuedPool` is used (by default), and it's working fine.

README.md
```diff
@@ -167,6 +167,14 @@ Note that execution options are applied lazily - on the `execute()` method call,
 
 ReadOnly/ReadWrite mode of a connection can't be changed while a transaction is in progress - first you must commit or rollback it.
 
+### Stale reads 
+To use the Spanner [Stale Reads](https://cloud.google.com/spanner/docs/reads#perform-stale-read) with SQLAlchemy you can tweak the connection execution options with a wanted staleness value. For example:
```
larkee, line 171: Read timestamp and minimum read timestamp expect a `datetime.datetime` object.
Maximum staleness and exact staleness expect a `datetime.timedelta` object.
This should probably be documented here in some way, possibly with examples for each, WDYT?

README.md
```diff
@@ -167,6 +167,14 @@ Note that execution options are applied lazily - on the `execute()` method call,
...
+```python
+with engine.connect().execution_options(staleness={"max_staleness": 5}) as connection:
```
larkee, line 173: Staleness are only supported for read-only transactions:
```suggestion
with engine.connect().execution_options(read_only=True, staleness={"max_staleness": datetime.timedelta(seconds=5)}) as connection:
```

test/test_suite.py
```diff
@@ -1594,3 +1594,13 @@ def test_read_only(self):
         with self._engine.connect().execution_options(read_only=True) as connection:
             connection.execute(select(["*"], from_obj=self._table)).fetchall()
             assert connection.connection.read_only is True
+
+    def test_staleness(self):
+        with self._engine.connect().execution_options(
+            staleness={"max_staleness": 5}
```
larkee, line 1600:
```suggestion
            read_only=True,
            staleness={"max_staleness": datetime.timedelta(seconds=5)}
```

test/test_suite.py
```diff
@@ -1594,3 +1594,13 @@ def test_read_only(self):
...
+        ) as connection:
+            connection.execute(select(["*"], from_obj=self._table)).fetchall()
+            assert connection.connection.staleness == {"max_staleness": 5}
```
larkee, line 1603:
```suggestion
            assert connection.connection.staleness == {"max_staleness": datetime.timedelta(seconds=5)}
```

README.md
IlyaFaer: Okay, added examples for all of them

larkee (approved):

PR #148, opened 2021-11-17
ci: remove unneeded presubmit configs

larkee:


skuruppu (approved):

skuruppu:
I think #150 will fix the issues with the system-3.8 tests. I haven't yet decided if we want to keep or remove the docs-presubmit.

PR #150, opened 2021-11-17
test: run compliance tests separately

skuruppu:

PR #151, opened 2021-11-18
test: separate GH actions

skuruppu:


skuruppu:
@IlyaFaer will you be able to help me figure out why the unit tests fail on GH Actions?
```
INTERNALERROR> configparser.NoSectionError: No section: 'db'
```

This happens regardless of whether separate the tests across runs or if you run it all togeter (see the failure in #150). AFAICT, we should never need to have the configparser do anything for the unit tests.

IlyaFaer:
@skuruppu, this is something SQLAlchemy is doing. There is a config file, where a database URL is set, right now we're creating it dinamically with `create_test_database.py`, here:
https://github.com/googleapis/python-spanner-sqlalchemy/blob/a64644b1a36fd10c57ddac5a15105d1a412ffbde/create_test_database.py#L91-L99

This config file will be used somewhere under the hood by SQLAlchemy.

So, speaking about the tests, I guess the fix is to add such a line:
```python
```
Into the `unit` session.

skuruppu:
Ah ok, thanks a lot @IlyaFaer. That was indeed the issue.

noxfile.py
```diff
@@ -133,6 +133,7 @@ def unit(session):
     session.install("opentelemetry-api==1.1.0")
     session.install("opentelemetry-sdk==1.1.0")
     session.install("opentelemetry-instrumentation==0.20b0")
+    session.run("python", "create_test_database.py")
```
larkee, line 136: I really don't like that we have to create a database to run the unit tests. Is it worth separating out the logic for dynamically setting the config into another script?

So we have a new script that defines setting the config
````
def set_config(instance_id):
  ...
  
if __name__ == "__main__":
   main(sys.argv[1:])
````
And then the `create_test_database.py` script can import and use the function:
```
from new_script import set_config
```

WDYT?

**Review by larkee — COMMENTED** [review]
LGTM other than a small nit regarding the unit test setup

File: noxfile.py (hunk 1 above)
Comment (skuruppu) [reply noxfile.py]: Yeah I agree, I think that's the right way to clean it up. PTAL.

**Review by larkee — APPROVED** [review]
LGTM

File: .github/workflows/test_suite.yml [hunk 2]
```diff
@@ -5,7 +5,40 @@ on:
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install nox
+        run: python -m pip install nox
+      - name: Run Unit Tests
+        run: nox -s unit
+        env:
+          SPANNER_EMULATOR_HOST: localhost:9010
+          GOOGLE_CLOUD_PROJECT: appdev-soda-spanner-staging
```
skuruppu, line 39: @larkee let me know if you have a different solution to this. The problem is that the unit tests are not really unit tests. It tries to create an actual Python client underneath, which requires `GOOGLE_APPLICATION_CREDENTIALS` to be set. We have nothing meaningful to set this env var to so the next best thing I could think of was to set the emulator host so that the client defaults to setting anonymous creds.

PR #152, opened 2021-11-18
test: run snippets tests on Kokoro

skuruppu:


skuruppu:
The snippets test did succeed, but one of the transactions was aborted in the compliance test so I'm rerunning it now.

larkee (commented):
If they must be run against production then we should leave the `GOOGLE_APPLICATION_CREDENTIALS` check in.

skuruppu:
> If they must be run against production then we should leave the `GOOGLE_APPLICATION_CREDENTIALS` check in.

Ah right but it turns out that Kokoro doesn't set up that variable. We were only doing that in CircleCI so the tests kept getting skipped on Kokoro. This now makes the tests run on Kokoro.

larkee:
> > If they must be run against production then we should leave the `GOOGLE_APPLICATION_CREDENTIALS` check in.
> 
> Ah right but it turns out that Kokoro doesn't set up that variable. We were only doing that in CircleCI so the tests kept getting skipped on Kokoro. This now makes the tests run on Kokoro.

Hmmm, how strange.. In that case, LGTM :+1:

larkee (approved):

PR #159, opened 2021-11-30
fix: raise Unimplemented error when creating temporary tables

IlyaFaer:


google/cloud/sqlalchemy_spanner/requirements.py
```diff
@@ -68,6 +68,11 @@ def isolation_level(self):
     def sequences(self):
         return exclusions.closed()
 
+    @property
+    def temporary_tables(self):
+        """target database supports temporary tables"""
```
skuruppu, line 73:
```suggestion
        """Target database supports temporary tables."""
```

google/cloud/sqlalchemy_spanner/requirements.py
```diff
@@ -68,6 +68,11 @@ def isolation_level(self):
...
+        return exclusions.closed()
```
skuruppu, line 74: Please also update the README to say that it's no longer supported.

skuruppu (changes requested):

google/cloud/sqlalchemy_spanner/requirements.py
larkee: The README already [states](https://github.com/googleapis/python-spanner-sqlalchemy/blob/a64644b1a36fd10c57ddac5a15105d1a412ffbde/README.md?plain=1#L198):
```
Temporary tables are not supported, real tables are used instead.
```

larkee (commented):
I did some investigating and it looks like if we modify `_spanner_temp_table_keyword_args` to return `{"prefixes": ["TEMPORARY"]}` then we should be able to differentiate between real and temporary tables in `post_create_table` by checking for the `TEMPORARY` prefix. This means we may still be able to raise an `Unimplemented` error.

From local testing, this seems to prevent creating a temporary table via the `TEMPORARY` prefix when declaring the table. However, the test suite started having failures which I'm guessing are from tests using temporary tables. This may mean more overrides of `define_tables` would be necessary.

@skuruppu WDYT?

skuruppu:
> I did some investigating and it looks like if we modify `_spanner_temp_table_keyword_args` to return `{"prefixes": ["TEMPORARY"]}` then we should be able to differentiate between real and temporary tables in `post_create_table` by checking for the `TEMPORARY` prefix. This means we may still be able to raise an `Unimplemented` error.
> 
> From local testing, this seems to prevent creating a temporary table via the `TEMPORARY` prefix when declaring the table. However, the test suite started having failures which I'm guessing are from tests using temporary tables. This may mean more overrides of `define_tables` would be necessary.
> 
> @skuruppu WDYT?

Yeah I think ideally we would prevent anyone from attempting to create temporary tables since they won't behave the way a user might expect when using real tables. So returning UNIMPLEMENTED is ideal if it's not too complicated to implement.

As for tests, how many tests will fail if we go down this path? Would it mean that we have to skip a large portion of the test suite or is it just a few tests? If it's a few, can we override the tests to create real tables?

google/cloud/sqlalchemy_spanner/requirements.py
skuruppu: Oh right, I meant that we should just say "Temporary tables are not supported." so that users don't attempt to use this as a feature.

google/cloud/sqlalchemy_spanner/requirements.py
IlyaFaer: Oh, yeah. I've copy pasted the comment from the original SQLAlchemy method.

IlyaFaer:
@skuruppu, I'd expect about 60% of tests failing. It's about 120 test cases.
Other don't use tables, just work with the dialect itself.

IlyaFaer:
@skuruppu, @larkee, yes, this idea with catching prefixes in `post_create_table` works fine. Only one test case had to be updated (reflection - a very big bunch of tests, which gives 40% of the whole suite). Pushing...

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -341,6 +341,9 @@ def post_create_table(self, table):
         cols = [col.name for col in table.primary_key.columns]
         post_cmds = " PRIMARY KEY ({})".format(", ".join(cols))
 
+        if table._prefixes == ["TEMPORARY"]:
```
larkee, line 344: nit: prefer to check if it's contained in the list in case other prefixes are used as well

```suggestion
        if "TEMPORARY" in table._prefixes:
```

larkee (approved):

skuruppu (approved):

PR #161, opened 2021-12-02
test: increase the package testing coverage

IlyaFaer:


google/cloud/sqlalchemy_spanner/requirements.py
```diff
@@ -16,7 +16,7 @@
 from sqlalchemy.testing.requirements import SuiteRequirements
 
 
-class Requirements(SuiteRequirements):
+class Requirements(SuiteRequirements):  # pragma: no cover
```
IlyaFaer, line 19: This class marks what features are supported by the dialect and which are not. Some of its marker-methods are used by SQLAlchemy to choose a right way to process a request, but some exist only as an information for user (and are not called by any methods). Thus, I think we can skip it in coverage.

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -373,10 +373,14 @@ def visit_TEXT(self, type_, **kw):
...
-    def visit_BINARY(self, type_, **kw):
+    def visit_BINARY(self, type_, **kw):  # pragma: no cover
+        """
+        The BINARY type is superseded by large_binary in
+        newer versions of SQLAlchemy (>1.4).
+        """
         return "BYTES({})".format(type_.length or "MAX")
 
-    def visit_large_binary(self, type_, **kw):
+    def visit_large_binary(self, type_, **kw):  # pragma: no cover
```
IlyaFaer, line 383: Here we have two data types. Binary is older and it's used before SQLAlchemy 1.4 version. Large binary is a newer type, which is used by new SQLAlchemy versions. As we're supporting both version ranges, we'll always get in a situation when one type was used, but not another one. As this will influence coverage percent, I'd skip them both in report.

google/cloud/sqlalchemy_spanner/provision.py
```diff
@@ -15,6 +15,6 @@
 from sqlalchemy.testing.provision import temp_table_keyword_args
 
 
-@temp_table_keyword_args.for_db("spanner")
+@temp_table_keyword_args.for_db("spanner")  # pragma: no cover
 def _spanner_temp_table_keyword_args(cfg, eng):
     return {"prefixes": ["TEMPORARY"]}
```
IlyaFaer, line 20: This function is dynamically registered to the dialect with proxies, so its influence is kinda "from the side". `coverage` little bit losing this part in the report, considering it as never executed during tests.

IlyaFaer:
Missings on lines 925, 938 are related to `Alembic` and separate nox session `migration_test`. Not sure if we can compile `coverage` reports from two different sessions with such a GH-workflow.

IlyaFaer:
912-917 missings are related to our override for `do_execute_no_params()` method. It was overriden only to add tracing calls there - no functionality changes, so this missing doesn't look very meaningful.

IlyaFaer:
As for the missing on the line 235, it's related to converting data to text. Looking at code, I'd say the only way how we can meaningfully cover this part, is to use table comments (which are not supported in Spanner) - in all the other cases (LIKE and ILIKE statements) the data type argument of the function is hardcoded and it's just a text string, so it's not gonna ever fail (because there is always a literal processor for text). I'd just skip this missing in coverage instead of creating an exotic mocks and non-existing situations.

skuruppu (approved):

PR #165, opened 2021-12-08
chore: setup release 1.0.0

skuruppu:
Release-As: 1.0.0

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -14,9 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-[egg_info]
-tag_build = dev
-
 [tool:pytest]
 addopts= --tb native -v -r fxX --maxfail=25 -p no:warnings
 python_files=test/*test_*.py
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -14,9 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-[egg_info]
-tag_build = dev
-
```
skuruppu, line 19: @IlyaFaer what is this for and is it ok that I removed it? I don't want this to be marked as a dev build (or for the versions to have any suffix).

IlyaFaer: Yeah, it can be deleted

IlyaFaer (approved):
I suppose now we can erase these lines from the readme:
Безымянный
Otherwise LGTM!

skuruppu:
> I suppose now we can erase these lines from the readme: Безымянный Otherwise LGTM!

Great catch @IlyaFaer. Thanks!

larkee (approved):

PR #180, opened 2021-12-23
docs: add a README section for the autoincremented ids

### LINKED ISSUE #176: Support autoincrementing integer IDs in some capacity
**Is your feature request related to a problem? Please describe.**
Autoincrementing serogate integer surrogate IDs are common among databases over the decades, and strongly recommended against within Cloud Spanner (https://cloud.google.com/spanner/docs/whitepapers/optimizing-schema-design [1]). I realize this is *not* an optimal solution, but it is a very compatible solution when schema migration is difficult for workloads with low latency / throughput requirements (this is python afterall).

Cloud spanner recommends a lot of application side boilerplate for generating autoincrementing IDs: https://cloud.google.com/solutions/sequence-generation-in-cloud-spanner#sequence_generators_using_database_table_rows [2]. TL;DR: Select the highest ID, increment, and insert all within a transaction is the simplest, most robust, and slowest option. I predict a lot of users migrating workloads to spanner will run into this issue and have to write their own code.

**Describe the solution you'd like**
sqlalchemy supports `sa.Column('col1', sa.Integer, autoincrement=True)`, but this is ignored by python-spanner-sqlalchemy (and an Error is raised during insertion that 'col1' was not supplied). Another option is `sa.Column('col1', sa.Integer, sa.Sequence('col1_auto_inc'))`, but table creation fails for lack of available sequences

Note: Both these tests were performed on the spanner emulator, not the actual thing. These might be emulator errors?

I'd like python-spanner-sqlalchemy to support either of the above solutions, or something similar.

**Describe alternatives you've considered**
The workaround is using a transaction, selecting the highest id, incrementing, and inserting within the same transaction (as listed in [2]). This requires reinventing the wheel for every insert for every user and application. It's not horrible, but it seems like adding it to the library would be ergonomic

IlyaFaer:
Closes #176

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,18 @@ eng = create_engine("spanner:///projects/project-id/instances/instance-id/databa
 autocommit_engine = eng.execution_options(isolation_level="AUTOCOMMIT")
 ```
 
+### Autoincremented IDs  
+Cloud Spanner doesn't support autoincremented IDs mechanism due to performance reasons ([see for more details](https://cloud.google.com/spanner/docs/schema-design#primary-key-prevent-hotspots)). Though it's not encouraged to do so, in case you *need* the feature, you can simulate it manually, for example:
+```python
+with engine.begin() as connection:
+    top_id = connection.execute(
+        select([user.c.user_id]).order_by(user.c.user_id.desc()).limit(1)
+    ).fetchone()
+    next_id = top_id[0] + 1 if top_id else 1
+
+    connection.execute(user.insert(), {"user_id": next_id})
+```
+
 ### Query hints  
 Spanner dialect supports [query hints](https://cloud.google.com/spanner/docs/query-syntax#table_hints), which give the ability to set additional query execution parameters. Usage example:
 ```python


README.md
```diff
@@ -157,6 +157,18 @@ eng = create_engine("spanner:///projects/project-id/instances/instance-id/databa
 autocommit_engine = eng.execution_options(isolation_level="AUTOCOMMIT")
 ```
 
+### Autoincremented IDs  
+Cloud Spanner doesn't support autoincremented IDs mechanism due to performance reasons ([see for more details](https://cloud.google.com/spanner/docs/schema-design#primary-key-prevent-hotspots)). Though it's not encouraged to do so, in case you *need* the feature, you can simulate it manually, for example:
```
skuruppu, line 161:
```suggestion
Cloud Spanner doesn't support autoincremented IDs mechanism due to performance reasons ([see for more details](https://cloud.google.com/spanner/docs/schema-design#primary-key-prevent-hotspots)). Though it's not encouraged to do so, in case you *need* the feature, you can simulate it manually as follows:
```

skuruppu (approved):

PR #189, opened 2022-01-10
feat: drop read_only on a connection returned back to a pool

IlyaFaer:
Current connections reset mechanism doesn't drop `read_only` property of a connection returned back to a connections pool. It makes the following cases possible:

- user initiates a `ReadOnly` connection, executes some operations and returnes the connection back to the pool (automatically)
- user initiates the next connection, without making it `ReadOnly`, but the pool returns the connection, which was toggled to the `ReadOnly` mode earlier (as it's a pool, it returns the connections created earlier, it is its purpos).

Such a situation can be very unexpected and will require for user to write additional logic to understand if the connection can be used in their case. To avoid it, we better drop `read_only` attribute to its default value when a connection is returned back to pool.

P.S. because of the same reasons we're already dropping `staleness` attribute on a connection reset.


google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -51,6 +51,9 @@ def reset_connection(dbapi_conn, connection_record):
     if getattr(dbapi_conn.connection, "staleness", None) is not None:
         dbapi_conn.connection.staleness = None
 
+    if getattr(dbapi_conn.connection, "read_only", None) is not None:
+        dbapi_conn.connection.read_only = False
+
```
IlyaFaer, line 56: Dropping the attribute on a connection reset

test/test_suite.py
```diff
@@ -1579,45 +1579,32 @@ def test_user_agent(self):
...
+        self._table = Table(
             "execution_options",
-            cls._metadata,
+            metadata,
             Column("opt_id", Integer, primary_key=True),
             Column("opt_name", String(16), nullable=False),
         )
 
-        cls._metadata.create_all(cls._engine)
+        metadata.create_all(self._engine)
```
IlyaFaer, line 1599: Splitting the test case into two different cases to avoid some emulator related issues

test/test_suite.py
```diff
@@ -1645,6 +1632,43 @@ def test_offset_only(self):
...
+
+        self._table = Table(
+            "execution_options",
+            metadata,
+            Column("opt_id", Integer, primary_key=True),
+            Column("opt_name", String(16), nullable=False),
+        )
+
+        metadata.create_all(self._engine)
+        time.sleep(1)
```
IlyaFaer, line 1653: As we're using `staleness`, and the emulator works really fast, we're getting in a situation when there is no test table yet, but we're already trying to read it, getting an error. Doing a small pause to avoid it.

vi3k6i5 (approved):
LGTM

PR #196, opened 2022-01-24
build(python): switch back to keystore for publication

parthea:
This PR pulls in changes from https://github.com/googleapis/synthtool/pull/1327 as we have migrated from secret manager to keystore for publishing to PyPI.


larkee (approved):
LGTM, thanks!

PR #198, opened 2022-01-27
feat: allow SQLAlchemy 1.4 use

IlyaFaer:
Allow SQLAlchemy 1.4 to be used with the dialect.

### DIFF
--- a/google/cloud/sqlalchemy_spanner/requirements.py
+++ b/google/cloud/sqlalchemy_spanner/requirements.py
@@ -29,6 +29,7 @@ class Requirements(SuiteRequirements):  # pragma: no cover
     def computed_columns_stored(self):
         return exclusions.open()
 
+    @property
     def sane_rowcount(self):
         return exclusions.closed()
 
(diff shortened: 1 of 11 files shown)


ansh0l:
Hi @larkee, I'm adding you as a reviewer, please have a look at if you get some time this week.

ansh0l (approved):
PR looks good to me, please resolve conflicts, and check with @asthamohta on the merge.

IlyaFaer:
Still trying to fix tests here. I've figured out that if I move a test file to the root folder, it's visible for `pytest`. But in `test` folder it always gives `0 collected`. So, problem is somehow related to the folder name/location, it's only left to understand what's eactly wrong.

asthamohta:
> Still trying to fix tests here. I've figured out that if I move a test file to the root folder, it's visible for `pytest`. But in `test` folder it always gives `0 collected`. So, the problem is somehow related to the folder name/location, it's only left to understand what's exactly wrong.

Thanks Ilya for the update

IlyaFaer:
@asthamohta, only this bizarre error left with SQLAlchemy 1.3.

I'm thinking what if we'll just drop tests for SQLAlchemy 1.3? Right now new changes for SQLAlchemy 1.3 are pretty rare and they are only about fixing bugs, not functionality. I suspect we have problems here just because 1.3 is not very supported already, so the old test suite wasn't properly updated for the new `pytest` versions, that's why 1.4 tests are working fine, while 1.3 are not collected (seems like it's happening because of `__init__` methods of the test cases). Maybe let's simply drop the 1.3 test session and stick to 1.4? Erasing the double of the test suite for 1.3 should make our life easier.

asthamohta (approved):

PR #220, opened 2022-06-13
fix: incorrect DDL generated when using server_default (#209)

### LINKED ISSUE #209: Incorrect DDL generated when using server_default
#### Environment details

  - Programming language: Python 3.10
  - OS: Windows 10
  - Package version: 1.0.0

#### Steps to reproduce
```python
from sqlalchemy import (
    Column,
    Integer,
    MetaData,
    String,
    Table,
    create_engine,
)

engine = create_engine(
    URI_STRING
)
metadata = MetaData(bind=engine)

user = Table(
    'users',
    metadata,
    Column('user_id', Integer, primary_key=True),
    Column('foo', String, server_default='bar'),
)

metadata.create_all(engine)
```

This generates the following DDL:
```sql
CREATE TABLE users (
        user_id INT64 NOT NULL,
        foo STRING(MAX) DEFAULT 'bar'
) PRIMARY KEY (user_id)
```
However, spanner syntax requires the default values to be wrapped in parenthesis like so:
```sql
CREATE TABLE users (
        user_id INT64 NOT NULL,
        foo STRING(MAX) DEFAULT ('bar')
) PRIMARY KEY (user_id)
```

Additionally, when specifying numeric types, the server_default argument (which must be passed in as a string) is still wrapped in quotation marks inside the DDL.

trevor-pope:
Added parenthesis to enclose DEFAULT values of CREATE TABLE statements, and changed the order of NOT NULL and DEFAULT statements to match the specifications of Spanner.

Fixes #209 

Could also probably add some tests for server_default.

I also realized that the intended way is to use sqlalchemy.text() to render default values, so the other problem mentioned at the bottom of #209 is not true.


IlyaFaer:
@asthamohta, okay, I've fixed everything here.

It appears now linter is checking `long_description` of the package. In the main `python-spanner` repository the README file is [used as a long description](https://github.com/googleapis/python-spanner/blob/3867882a14c9a2edeb4a47d5a77ec10b2e8e35da/setup.py#L76).

Also it requires to use `.rst` format README file instead of GitHub default `.md` (otherwise `list_setup_py` will still fail). So, I've converted the README file as well.

Let's merge the PR, and I will add tests for it separately (not very convenient to work in client's fork).

asthamohta (approved):

PR #234, opened 2022-08-25
fix: update dialect name for ALTER operation overrides

### LINKED ISSUE #232: NOT NULL statements are not working for generated Alembic migrations
the following 2 statements in the Spanner DDL window do the trick :
ALTER TABLE test ADD COLUMN test STRING(MAX);
ALTER TABLE test ALTER COLUMN test STRING(MAX) NOT NULL;

But there is no native way of doing it in SQLAlchemy
The workaround we currently use is to manually edit the migration file to add the two above statements using op.execute(). Tt would be nice if it worked out-of-the-box

IlyaFaer:
Fixes #232

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -65,7 +65,14 @@ UPGRADE_CODE = """def upgrade():
         'account',
         'name',
         existing_type=sa.String(70),
-    )"""
+    )
+    op.alter_column(
+        'account',
+        'description',
+        existing_type=sa.Unicode(200),
+        nullable=False,
+    )
+    """
 
 
 BLACK_VERSION = "black==22.3.0"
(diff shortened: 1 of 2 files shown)


noxfile.py
```diff
@@ -65,7 +65,14 @@ class = StreamHandler
...
         existing_type=sa.String(70),
-    )"""
+    )
+    op.alter_column(
+        'account',
+        'description',
+        existing_type=sa.Unicode(200),
+        nullable=False,
+    )
+    """
```
IlyaFaer, line 75: These strings are used in migration tests as a migration script.
Adding an operation, which was failing in the original issue, to make sure the case is tested.

noxfile.py
```diff
@@ -254,14 +261,14 @@ def migration_test(session):
...
+    with open(files[0], "rb") as f:
+        script_code = f.read().decode()
 
     script_code = script_code.replace(
         """def upgrade() -> None:\n    pass""", UPGRADE_CODE
     )
-    with open(files[0], "w") as f:
-        f.write(script_code)
+    with open(files[0], "wb") as f:
+        f.write(script_code.encode())
```
IlyaFaer, line 271: Locally I've got an error with rewriting a migration script by a program - some symbols were missing. Rewriting the file, using bytes, fixes the problem.

asthamohta (approved):

Issue #231, opened 2022-08-18
Introspection method returns primary key indexes, making Alembic migrations inconvenient


IlyaFaer:
When we generate migrations with Alembic, it always tries to drop a bunch of indexes. Running the migration fails because they are used by foreign keys.

The dropped indexes are those for the primary key of each table. Those indexes are never explicitly created by Alembic, but they are implicitly by virtue of being on primary key columns. This might explain why Alembic tries to drop them.
Right now, we delete those drop_index manually every time we generate a migration, which is tedious

IlyaFaer:
@asthamohta, @ansh0l

IlyaFaer:
So, the proposed idea was that Alembic is trying to drop primary keys of tables (while primary keys should be managed automatically by Spanner and never touched by users). To build a migration, Alembic uses an introspection method to understand what is the table structure and what needs to be done. All of it is happening here:

https://github.com/sqlalchemy/alembic/blob/873027160f88af277b73223e9fe33bfba315532f/alembic/autogenerate/compare.py#L570

https://github.com/sqlalchemy/sqlalchemy/blob/f8c4dba4e9f130c18ce00597c036bc26ae7abf90/lib/sqlalchemy/engine/reflection.py#L1135

And the dialect introspection method is already ignoring primary key indexes:
https://github.com/googleapis/python-spanner-sqlalchemy/blob/8bab5fb33cb7d08905b5ec5c17ee8f26b848b359/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py#L691

Thus, it seems Alembic should not try to delete primary key indexes, as it'll not see them in an existing table.

I've tried to reproduce the case by creating two tables, one of which is foreign keyed into another, and introspection works fine. I even tried to delete the tables (when it's done with SQLAlchemy, it first drops the table indexes - otherwise Spanner will not allow to drop the table, breaking half of SQLAlchemy functionality). Here are codes:

```sql
CREATE TABLE Singers (
            SingerId     INT64 NOT NULL,
            FirstName    STRING(1024),
            Params       ARRAY<JSON>,
        ) PRIMARY KEY (SingerId)
```

```sql
CREATE TABLE Song (
        SongId INT64 NOT NULL,
        Name STRING(256),
        Author INT64 NOT NULL,
        CONSTRAINT FK_SongAuthor FOREIGN KEY (Author) REFERENCES Singers (SingerId)
    ) PRIMARY KEY (SongId)
```

[...]

PR #241, opened 2022-08-31
fix: Spanner auto managed indexes should not be introspected

IlyaFaer:
Towards #231 

SQLAlchemy introspection method returns all the indexes, including those managed by Spanner internally. As any manipulations with indexes auto managed by Spanner can cause troubles, it's fair to prevent dialect from noticing these indexes. Fortunately, `information_schema` includes a flag, which makes it easy.

### DIFF
--- a/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
+++ b/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
@@ -691,6 +691,7 @@ WHERE
     AND i.index_type != 'PRIMARY_KEY'
     AND i.spanner_is_managed = FALSE
 GROUP BY i.index_name, i.is_unique
+ORDER BY i.index_name
 """.format(
             table_name=table_name
         )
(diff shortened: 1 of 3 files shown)


google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -689,7 +689,9 @@ def get_indexes(self, connection, table_name, schema=None, **kw):
 WHERE
     i.table_name="{table_name}"
     AND i.index_type != 'PRIMARY_KEY'
+    AND i.spanner_is_managed = FALSE
```
IlyaFaer, line 692: Don't show automatic indexes ruled by Spanner under the hood

test/test_suite_13.py
```diff
@@ -939,6 +939,13 @@ def test_array_reflection(self):
 
         tab.drop()
 
+    def _assert_insp_indexes(self, indexes, expected_indexes):
+        expected_indexes.sort(key=lambda item: item["name"])
+
+        index_names = [d["name"] for d in indexes]
+        exp_index_names = [d["name"] for d in expected_indexes]
+        assert sorted(index_names) == sorted(exp_index_names)
```
IlyaFaer, line 947: Overriding the test to check indexes in ordered fashion

ansh0l (approved):

PR #250, opened 2022-09-27
fix: adding requirements

asthamohta:
Adding requirements.in and .txt with hashes


IlyaFaer:
@asthamohta, so, it looks like `keyring` package requires `jeepney`, and it's version is not particular.
```
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    jeepney>=0.4.2 from https://files.pythonhosted.org/packages/ae/72/2a1e2290f1ab1e06f71f3d0f1646c9e4634e70e1d37491535e19266e8dc9/jeepney-0.8.0-py3-none-any.whl#sha256=c0a454ad016ca575060802ee4d590dd912e35c122fa04e70306de3d076cce755 (from keyring==23.9.3->-r .kokoro/requirements.txt (line 280))
```
I suppose we can simply add the `jeepney` package into requirements to explicitly set its versions.

IlyaFaer (approved):

PR #270, opened 2022-10-13
Fix alembic incompatibility with sqlalchemy < 1.3.11

### LINKED ISSUE #269: alembic incompatibility with sqlalchemy < 1.3.11 due to computed columns
#### Environment details

  - Programming language: python
  - OS: macos
  - Language runtime version: 3.7.14 
  - Package version: sqlalchemy-spanner 1.3.2, sqlalchemy 1.2.12, alembic 0.9.3

#### Steps to reproduce

  1. install any sqlalchemy version under 1.3.11
  2. run any alembic migration 
 
Easy way to try this using the existing tests--in the [migration_test function](https://github.com/googleapis/python-spanner-sqlalchemy/blob/main/noxfile.py#L225) in noxfile.py, add 

```session.run("pip", "install", "sqlalchemy>=1.1.13,<=1.3.10", "--force-reinstall")```

 at the beginning after the imports to force install the sqlalchemy versions with the incompatibilities. Then run the tests using `nox -s migration_test`. In `alembic upgrade head`, and the following error should occur:

```
<removed>
  File "google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py", line 352, in get_column_specification
    if column.computed is not None:
  File "nox/migration_test/lib/python3.8/site-packages/sqlalchemy/sql/elements.py", line 690, in __getattr__
    key)
AttributeError: Neither 'Column' object nor 'Comparator' object has an attribute 'computed'
```

This is because the sqlalchemy_spanner alembic code [here](https://github.com/googleapis/python-spanner-sqlalchemy/blob/main/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py#L352-L353) refers to the `computed` attribute of `Column`, which was introduced in sqlalchemy 1.3.11 ([release notes](https://github.com/sqlalchemy/sqlalchemy/releases/tag/rel_1_3_11), [code ref](https://github.com/sqlalchemy/sqlalchemy/compare/rel_1_3_10...rel_1_3_11#diff-04bfd5c1f4dc2b278cac4df823a4dca6fbb03a1c16d8dc7bb88714cf3858bce3R1347)). So, for sqlalchemy versions < 1.3.11, this will throw the `AttributeError` above since `Column`s don't have the computed attribute. 

A simple but maybe hacky fix would be to just check if the attribute exists before accessing it:

New version
```
if hasattr(column, "computed") and column.computed is not None:
    colspec += " " + self.process(column.computed)
```

Old version
```
if column.computed is not None:
    colspec += " " + self.process(column.computed)
```

but definitely open to suggestions for how to fix this!

elainearbaugh:
Also adds a unit test specifically targeting these sqlalchemy versions (this test fails without the fix). Not sure if this is the best way to check for this attribute so lmk if there's a more elegant way to do it

Fixes #269

- [x] Tests pass
- [n/a] Appropriate changes to README are included in PR

### DIFF
--- a/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
+++ b/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
@@ -349,7 +349,7 @@ class SpannerDDLCompiler(DDLCompiler):
         if default is not None:
             colspec += " DEFAULT (" + default + ")"
 
-        if column.computed is not None:
+        if hasattr(column, "computed") and column.computed is not None:
             colspec += " " + self.process(column.computed)
 
         return colspec
--- a/noxfile.py
+++ b/noxfile.py
@@ -223,6 +223,19 @@ def unit(session):
 
 @nox.session(python=DEFAULT_PYTHON_VERSION)
 def migration_test(session):
+    """Test migrations with SQLAlchemy v1.3.11+ and Alembic"""
+    session.run("pip", "install", "sqlalchemy>=1.3.11", "--force-reinstall")
+    _migration_test(session)
+
+
+@nox.session(python=DEFAULT_PYTHON_VERSION)
+def migration_test_1310(session):
+    """Test migrations with SQLAlchemy 1.3.10 or lower and Alembic"""
+    session.run("pip", "install", "sqlalchemy>=1.1.13,<=1.3.10", "--force-reinstall")
+    _migration_test(session)
+
+
+def _migration_test(session):
     """Migrate with SQLAlchemy and Alembic and check the result."""
     import glob
     import os


IlyaFaer:
@elainearbaugh, the PR looks okay to me. They only thing needed is for you to sign the CLA agreement:
https://github.com/googleapis/python-spanner-sqlalchemy/blob/main/CONTRIBUTING.md#contributor-license-agreement
Then we can merge it.

asthamohta (approved):

elainearbaugh:
@IlyaFaer thanks! I just got the CLA signed this week, working through the other issues now

elainearbaugh:
@IlyaFaer resolved the commit issues as well so if you could allow running workflows and re-review that'd be awesome! Sorry for being slow, it took a while for my org to get the CLA through

elainearbaugh:
I'm not sure what's going on with the Kokoro failure--seems like it hit a "ERROR: Aborting VM command due to timeout of 10800 seconds" (3 hours)? I added some additional testing but it definitely doesn't take 3 hours running locally...I don't think I have the ability to rerun the tests if it's a flake so not sure what my next steps should be here

olavloite:
Closing, as this seems to have been implemented through another PR (and this PR is also very old)

PR #309, opened 2023-03-02
feat: add support for leader aware routing

asthamohta:


noxfile.py
```diff
@@ -192,7 +192,7 @@ def compliance_test_14(session):
     session.install("-e", ".[tracing]")
     session.run("python", "create_test_database.py")
 
-    session.install("sqlalchemy>=1.4")
+    session.install("sqlalchemy>=1.4,<2.0")
```
surbhigarg92, line 195: Why is this, don't we support sqlalchemy 2.0 ?

test/test_suite_14.py
```diff
@@ -1767,6 +1767,26 @@ def test_user_agent(self):
...
+    """
+
+    def setUp(self):
+        self._engine = create_engine(
+            "spanner:///projects/appdev-soda-spanner-staging/instances/"
+            "sqlalchemy-dialect-test/databases/compliance-test"
+            "?route_to_leader_enabled=False"
+        )
+
+    def test_route_to_leader_enabled(self):
```
surbhigarg92, line 1783: Can we also add an test for default option, if nothing is passed then 
`connection.connection.instance._client.route_to_leader_enabled is True`

olavloite:
Tests currently fail due to https://github.com/googleapis/python-spanner/pull/1223

test/test_suite_14.py
olavloite: Added tests for enabled/disabled/default

PR #336, opened 2023-07-31
feat: support sequences

harshachinta:
This PR adds support for sequences in Spanner SQLAlchemy through Bit Reversed Sequences.


harshachinta:
@asthamohta Can you please review this?

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -1294,6 +1364,32 @@ def has_table(self, connection, table_name, schema=None, **kw):
...
+
+        Given a :class:`_engine.Connection` object and a string
+        `sequence_name`, return True if the given sequence exists in
+        the database, False otherwise.
+        """
+
+        with connection.connection.database.snapshot() as snap:
+            rows = snap.execute_sql(
+                """
+SELECT true
```
asthamohta, line 1379: nit: can we indent this properly

test/test_suite_13.py
```diff
@@ -2059,3 +2063,128 @@ def test_create_engine_wo_database(self):
...
+                Integer,
+                sqlalchemy.Sequence("noret_id_seq"),
+                primary_key=True,
+            ),
+            Column("data", String(50)),
+            implicit_returning=False,
+        )
+
+    def test_insert_roundtrip(self, connection):
+        connection.execute(self.tables.seq_pk.insert(), dict(data="some data"))
```
asthamohta, line 2112: I see from the class that this is inherited from the code is the exact same, why are we overriding the functions here then?

test/test_suite_13.py
```diff
@@ -2059,3 +2063,128 @@ def test_create_engine_wo_database(self):
...
+        self._assert_round_trip(self.tables.seq_pk, connection)
+
+    def test_insert_lastrowid(self, connection):
+        r = connection.execute(self.tables.seq_pk.insert(), dict(data="some data"))
+        assert len(r.inserted_primary_key) == 1
+        is_instance_of(r.inserted_primary_key[0], int)
+
+    def test_nextval_direct(self, connection):
+        r = connection.execute(self.tables.seq_pk.c.id.default)
```
asthamohta, line 2121: Does connection.scalar not work for sqlalchemy spanner??

test/test_suite_13.py
```diff
@@ -2059,3 +2063,128 @@ def test_create_engine_wo_database(self):
...
+        is_instance_of(r, int)
```
asthamohta, line 2122: Why are we not checking for equal value as expected here and only type?

harshachinta: The actual test validates `eq_(r, testing.db.dialect.default_sequence_base)`, where `testing.db.dialect.default_sequence_base` is the starting value and is normally fixed (for ex: `1` in most dialects).

In Spanner SQLAlchemy, we support sequences through bit reversal and hence this is never fixed and keep changing.
This restriction does not allow us to verify the value and instead verify if the returned values is an `int`.

test/test_suite_13.py
harshachinta: `connection.scalar()`  method is designed to be used with SQL expressions or queries executed using the execute() method. Here `self.tables.seq_pk.c.id.default` internally calls a query and hence cannot be used with `scalar()`.
Though when I  tried it was throwing the following error
`AttributeError: 'int' object has no attribute 'scalar'`.

I think the original test also does not use `scalar` and instead uses `execute()` because of this reason.

google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
harshachinta: Done

test/test_suite_13.py
harshachinta: When I was running on my local, with out overriding I used to get `Transaction was aborted` error and I was not able to understand why it was happening. When I wrote the same test case in this file it worked. So I didn't remove.

I removed it now and the tests seem to pass.

test/test_suite_13.py
asthamohta: Understood, thanks

test/test_suite_13.py
asthamohta: Can we verify that the values are incremental atleast, or something of those sorts?

asthamohta (approved):

test/test_suite_13.py
harshachinta: As discussed offline, there is no other alternative due to randomness.

PR #346, opened 2023-10-10
fix: db.params OpenTelemetry integration issue

### LINKED ISSUE #345: Integration between Spanner SqlAlchemy and OpenTelemetry is generating a lot of warnings
We have an issue where integration between this package and OpenTelemetry is generating a lot of warnings.

In google.cloud.sqlalchemy_spanner `sqlalchemy_spanner.py` when it executes `do_execute` it calls `trace_call` but before doing so it sets the `trace_attributes` including `db.params` which are the values of the SQL query.

In `trace_call`, it checks if `HAS_OPENTELEMETRY_INSTALLED` and if so it will add these `db.params` atttributes to the trace and performs the trace.

But, in the OpenTelemetry Python package, it will check all attributes. One of the check that is performed on the attributes is that if an attribute is a Sequence type, all elements of the sequence must be of the same type. If not, it will throw a warning. This checks happens in the Python OpenTelemetry (external to Google) package in file `attributes/__init__.py` in the function `_clean_attribute` (see checks and _logger.warning calls).

The issue is that the Google Spanner SQLAlchemy package will store it’s `db.params` in the various types that they represent (which is good on it's own). It will then send these to the OpenTelemetry package who will check they’re all of the same type and throw a warning if not. So for all SQL queries that contain multiple datatype, OpenTelemetry will generate a warning. 

This results in a huge number of these warnings in our GCP Log Explorer.
This could be fixed in a number of ways including sending these values as strings when sending them to OpenTelemetry.

I've made some tests and if I change this package `sqlalchemy_spanner/_opentelemetry_tracing.py` file to use the following code:

```
if extra_attributes:
        if isinstance(extra_attributes, dict):
            for k, v in extra_attributes.items():
                if isinstance(v, collections.abc.Sequence):
                    extra_attributes[k] = [str(e) for e in v]

        attributes.update(extra_attributes)
```

I'm no longer receiving any warnings as all sequences values are converted to string before tracing.

gillesleblanc:
When db.params are not all of the same type OpenTelemetry will log a warning.

Prevent this by stringifying values.

Also adds a  SQLALCHEMY_SPANNER_TRACE_HIDE_QUERY_PARAMETERS env var to prevent logging db.params to OpenTelemetry.

Fixes #345

### DIFF
--- a/google/cloud/sqlalchemy_spanner/_opentelemetry_tracing.py
+++ b/google/cloud/sqlalchemy_spanner/_opentelemetry_tracing.py
@@ -14,6 +14,9 @@
 
 """Manages OpenTelemetry trace creation and handling"""
 
+import collections
+import os
+
 from contextlib import contextmanager
 
 from google.api_core.exceptions import GoogleAPICallError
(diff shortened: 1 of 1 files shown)


olavloite:
@gillesleblanc Sorry for the slow turnaround on this PR. The build errors seem to be unrelated to this change. I'm working on fixing those in #351, and after that we should be able to move this one forward as well.

olavloite (approved):

PR #374, opened 2023-11-14
build: allow merging pull requests that are not up to date

olavloite:
Allow pull requests that are not up to date with main to be merged, as we otherwise have re-run all tests for each PR before merging, which takes too long.

### DIFF
--- a/.github/sync-repo-settings.yaml
+++ b/.github/sync-repo-settings.yaml
@@ -5,7 +5,7 @@ branchProtectionRules:
 # Defaults to `main`
 - pattern: main
   requiresCodeOwnerReviews: true
-  requiresStrictStatusChecks: true
+  requiresStrictStatusChecks: false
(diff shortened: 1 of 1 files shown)


harshachinta (approved):
Instead of relaxing strict status check altogether, should we consider removing the long running checks from `requiredStatusCheckContexts`? This way atleast the basic checks can be validated?

olavloite:
> @olavloite Instead of relaxing strict status check altogether, should we consider removing the long running checks from `requiredStatusCheckContexts`? This way atleast the basic checks can be validated?

Note that this change does not mean that the required checks do not need to pass. Only that the branch does not need to be up to date with all changes in main. That makes it easier to run the tests for multiple (dependency) pull requests in parallel, and then merge them all. That means that there is still a small possibility that different dependencies break each other, but that will then show up in the build of the main branch (and in any release pull request).

PR #385, opened 2024-02-05
fix: Table name should be quoted by back quotes (`) on DROP TABLE

ankiaga:
https://github.com/googleapis/python-spanner-sqlalchemy/issues/349

### DIFF
--- a/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
+++ b/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
@@ -438,7 +438,7 @@ class SpannerDDLCompiler(DDLCompiler):
         for index in drop_table.element.indexes:
             indexes += "DROP INDEX {};".format(self.preparer.quote(index.name))
 
-        return indexes + constrs + str(drop_table)
+        return indexes + constrs + super().visit_drop_table(drop_table)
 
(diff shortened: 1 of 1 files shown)


google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -438,7 +438,7 @@ def visit_drop_table(self, drop_table, **kw):
         for index in drop_table.element.indexes:
             indexes += "DROP INDEX {};".format(self.preparer.quote(index.name))
 
-        return indexes + constrs + str(drop_table)
+        return indexes + constrs + super().visit_drop_table(drop_table)
```
olavloite, line 441: Can we use the same strategy for quoting identifiers for all operations in this method (and probably also in the entire file). In this method, it seems that most identifiers are quoted using `preparer.quote(..)`.

Example: `"DROP INDEX {};".format(self.preparer.quote(index.name))`. Should we do that here as well?

ankiaga: `self.preparer.quote` (its a super class variable not overridden here) is the correct way as internally `super().visit_drop_table()` also calls `self.preparer.quote`

olavloite (approved):

PR #420, opened 2024-10-31
test: add system tests

olavloite:
1. Adds a system test session to the nox file + one (very simple) system test.
2. Adds a GitHub Actions workflow to run the system tests on the emulator.
3. Runs the system tests on Spanner in the Kokoro job.


olavloite:
cc @rayudu3745

noxfile.py
```diff
@@ -252,6 +252,41 @@ def compliance_test_20(session):
...
+        "pytest-cov",
+        "pytest-asyncio",
+    )
+
+    session.install("mock")
+    session.install(".[tracing]")
+    session.install("opentelemetry-api==1.27.0")
+    session.install("opentelemetry-sdk==1.27.0")
+    session.install("opentelemetry-instrumentation==0.48b0")
+    session.run("python", "create_test_database.py")
```
asthamohta, line 283: If my understanding is correct, we cleanup the databases only when a new system test is run, right? Do we want to keep the database around to check for errors? Shouldn't we do a cleanup post the system tests being run as well?

olavloite: Correct, we were only cleaning up during the following run (and then only databases older than 4 hours). I've added a cleanup job that is executed after the system tests have executed.

asthamohta (approved):

PR #533, opened 2024-12-03
docs: add sample for read-only transactions

### LINKED ISSUE #493: Read-only transaction support
Spanner SQLAlchemy + dbapi do not support read-only transactions. Read-only transactions are much more efficient for reading data than using a read/write transaction.

olavloite:
Adds a sample and documentation for read-only transactions.

Fixes #493


bhatt4982 (approved):
LGTM

PR #558, opened 2024-12-17
feat: support request and transaction tags

### LINKED ISSUE #525: Support Spanner Request Options Tags
Cloud Spanner API enables tags [request_tag / transaction_tag] for RequestOptions parameter in requests which used for statistics collection on spanner insights, and can be helpful on detecting anomalies and improving spanner database queries / transactions. 

[RequestOptions Reference](https://cloud.google.com/spanner/docs/reference/rest/v1/RequestOptions)

olavloite:
Do not merge: Depends on https://github.com/googleapis/python-spanner/pull/1262

Fixes #525


olavloite:
@ahr9n 

> BTW, this `request_tag = self.execution_options.get("request_tag")` means the request tag will be on connecting to spanner engines [so, it will be the same for all the transaction queries -- on transaction level], but it should be on executing queries [to be on query level], right?

You can set execution_options on a lower level than the engine. E.g. it can also be set for specific queries like this:

```python
with Session(engine.execution_options(read_only=True)) as session:
    # Execute two queries in a read-only transaction.
    session.scalars(
        select(Singer).execution_options(request_tag="my-tag-1")
    ).all()
    session.scalars(
        select(Singer).execution_options(request_tag="my-tag-2")
    ).all()
```

The request tag is then set on the `cursor` that is being used for the current query:

```python
request_tag = self.execution_options.get("request_tag")
if request_tag:
    self.cursor.request_tag = request_tag
```

bhatt4982 (commented):
LGTM

PR #639, opened 2025-04-22
feat: support schemas in queries and dml statements

waltaskew:
Add support for SELECT, UPDATE and DELETE statements against tables in
schemas.

Schema names are not allowed in Spanner SELECT statements. We need to
avoid generating SQL like

```sql
SELECT schema.tbl.id
FROM schema.tbl
```

To do so, we alias the table in order to produce SQL like:

```sql
SELECT tbl_1.id, tbl_1.col
FROM schema.tbl AS tbl_1


olavloite (approved):

olavloite:
@waltaskew Thank you for this great contribution!

waltaskew:
> @waltaskew Thank you for this great contribution!

Thanks for reviewing and accepting! Do you have any plans around when the next release of the library will be?

Issue #638, opened 2025-04-21
Schema Support


waltaskew:
I'm interested in using SQLAlchemy to interact with Spanner, but my organization is planning to organize tables into schemas. It looks like the dialect doesn't currently support interacting with tables in schemas.

Is this something folks have plans to take on? I'd be happy to help, especially if folks could point me in the right direction or share thoughts they have about how to implement this.

waltaskew:
Was actually able to make some progress on this today. Looks like the problem is that schema names aren't allowed in spots like SELECT bodies. Fortunately, there's an example in the mssql dialect which shows how to avoid them by aliasing the tables:

https://github.com/googleapis/python-spanner-sqlalchemy/pull/639

https://github.com/sqlalchemy/sqlalchemy/blob/801e1e1982ebc5ecd8d77d46615311c6c7dc9b94/lib/sqlalchemy/dialects/mssql/base.py#L710

PR #681, opened 2025-06-02
bug: Support Retrieval of Cross-Schema Foreign Keys

waltaskew:
Allow `get_multi_foreign_keys` to retrieve foreign keys between tables in different schemas. The current SQL used to retrieve them has a mistake in the JOINs used to retrieve them, confusing the `table_schema` column and the `constraint_schema` column. This goes unnoticed as long as everything is in the same schema.

fixes: #638


waltaskew:
@olavloite would you mind taking a look at this when you have a chance? We're having a bit of trouble from this bug.

olavloite (approved):

Issue #686, opened 2025-06-04
Enums Not Enforced Server-Side


waltaskew:
Currently, creating an Enum type with the dialect ala:

```python
sa.Column(
    "an_enum",
    sa.Enum("A", "B", "C", create_constraint=True),
),
```

creates a string column with no server-side checks on enum values. This appears to be because the `supports_native_enum` field on the dialect is currently [set to True](https://github.com/googleapis/python-spanner-sqlalchemy/blob/5077ddb7e71cf60bdb14cef8d34793ab94acb730/google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py#L801C5-L801C25), which [indicates to SQLAlchemy](https://docs.sqlalchemy.org/en/20/core/internals.html#sqlalchemy.engine.default.DefaultDialect.supports_native_enum) that it doesn't need to generate the `CHECK` constraints to actually enforce the enum values server-side. I think this should be set to False so we generate those `CHECK` constraints.

Spanner's ENUM looks to [depend on protos](https://cloud.google.com/spanner/docs/reference/standard-sql/data-types#enum_type) which seems harder to implement. I think just generating those `CHECK` constraints would be a helpful improvement!

PR #694, opened 2025-06-12
feat: Support Server-Side Checks for Enums

waltaskew:
Spanner uses protos for enums. Creating a column like Column("an_enum", Enum("A", "B", "C")) will result in a String column. Setting supports_native_enum to False means SQLAlchemy will generate check constraints to enforce the enum values server-side rather than rely on a the database's native enum type to enforce them.

Fixes: #686


google/cloud/sqlalchemy_spanner/sqlalchemy_spanner.py
```diff
@@ -798,10 +798,16 @@ class SpannerDialect(DefaultDialect):
...
     sequences_optional = False
     supports_identity_columns = True
-    supports_native_enum = True
     supports_native_boolean = True
     supports_native_decimal = True
     supports_statement_cache = True
+    # Spanner uses protos for enums. Creating a column like
+    # Column("an_enum", Enum("A", "B", "C")) will result in a String
+    # column. Setting supports_native_enum to False means SQLAlchemy
+    # will generate check constraints to enforce the enum values
```
olavloite, line 807: I'm not sure this change is enough to actually achieve this. Would you otherwise mind adding a test that shows that it does?

I tried the same here, but it does not seem to make any difference: https://github.com/googleapis/python-spanner-sqlalchemy/pull/698/files

aakashanandg (approved):

PR #770, opened 2025-10-08
build: update default python version to 3.12

olavloite:
- Remove all tests with Python 3.8 and replace them with test runs with 3.12.
- Remove all tests with SQLAlchemy 1.3, as that version is no longer supported.


bhatt4982 (approved):
LGTM...

PR #799, opened 2025-11-10
chore: exclude last_update from the THEN RETURN clause

olavloite:
Columns that are assigned a `PENDING_COMMIT_TIMESTAMP()` value cannot be
included in `THEN RETURN` clauses.

Fixes this build error: https://github.com/googleapis/python-spanner-sqlalchemy/actions/runs/19180600096/job/54948361746?pr=798

A recent fix in the Emulator caused the samples for SQLAlchemy to start fail. That is not due to a bug in the Emulator, but due to a bug __fix__ in the Emulator. Previously, the Emulator would allow something that would fail on Spanner. That difference has now been fixed.

### DIFF
--- a/samples/model.py
+++ b/samples/model.py
@@ -199,12 +199,16 @@ class TicketSale(Base):
         String(36), ForeignKey("singers.id", spanner_not_enforced=True)
     )
     # Create a commit timestamp column and set a client-side default of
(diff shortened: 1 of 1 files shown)


bhatt4982 (approved):
LGTM

